PR #127, opened 2023-09-29
feat: service worker

TorstenDittmann:


src/service-worker.js
```diff
@@ -0,0 +1,62 @@
...
+    ...build, // the app itself
+    ...files  // everything in `static`
+];
+
+self.addEventListener('install', (event) => {
+    // Create a new cache and add all files to it
+    async function addFilesToCache() {
+        const cache = await caches.open(CACHE);
+        await cache.addAll(ASSETS);
+    }
```
ArmanNik, line 17: Should we try catch this in case there are exceptions?

TorstenDittmann: I actually want to throw that error back up 👍🏻

TorstenDittmann: This is then handled by the browser

PR #180, opened 2023-10-03
feat: terms

ArmanNik:


src/routes/terms/+page.svelte
```diff
@@ -1,42 +1,453 @@
...
+                >
+                    <span class="aw-description">Table of contents</span>
+                    <span class="icon-menu-alt-4" aria-hidden="true" />
+                </button>
+            </header>
+            <aside class="aw-grid-120-1fr-auto-side aw-is-mobile-closed">
+                <div class="aw-page-steps">
+                    <div class="aw-page-steps-location aw-is-not-mobile" style="--location:0%;">
+                        <button class="aw-page-steps-location-button">
+                            <svg
```
TGlide, line 52: Prefer icon-font for icons or `img` for SVGs, for performance reasons

src/routes/terms/+page.svelte
```diff
@@ -1,42 +1,453 @@
...
+                                    <clipPath id="clip0_1684_10747">
+                                        <rect width="16" height="16" fill="white" />
+                                    </clipPath>
+                                </defs>
+                            </svg>
+                        </button>
+                    </div>
+                    <ul class="aw-page-steps-list aw-sub-body-500">
+                        <li>
+                            <a href="#introduction" class="is-selected">Introduction</a>
```
TGlide, line 136: This will always be selected this way, but then I noticed privacy is also this way... Will create an Asana task so I can fix it, you can leave it as is

TGlide:
@TorstenDittmann shouldn't we use markdoc for pages like this?

TorstenDittmann:
> @TorstenDittmann shouldn't we use markdoc for pages like this?

Probably, but that means adding a complete new layout for it.

TorstenDittmann (approved):

PR #185, opened 2023-10-03
add asset imgs & refactor footer nav

TGlide:
## What does this PR do?

- Adds assets imgs to assets page
- Refactors Footer Nav to use accessible accordions
- Adds dynamic nav indication to assets page
- Add link to assets page in footer


src/routes/assets/+page.svelte
```diff
@@ -1,648 +1,666 @@
...
+
+    $: progress = (function getProgress() {
+        const sections = Object.values(Section);
+        const index = sections.indexOf(selected);
+        return (index / (sections.length - 1)) * 100;
+    })();
+
+    const handleVisibility = (section: Section) => {
+        return (e: CustomEvent<boolean>) => {
+            console.log(selectedMap);
```
TorstenDittmann, line 45: leftover

src/routes/assets/+page.svelte
```diff
@@ -1,648 +1,666 @@
...
+    enum Section {
+        NAMING = 'Naming',
+        LOGOTYPE = 'Logotype',
+        LOGOMARK = 'Logomark',
+        COLORS = 'Brand colors',
+        VISUALS = 'Product visuals',
+        CONTACT = 'Contact us'
+    }
+
+    const getSectionId = (section: Section) => section.toLowerCase().replace(/\s/g, '-');
```
TorstenDittmann, line 22: fyi, this part we probably wanna re-use for pages like `/privacy`

TGlide: Agreed! I was thinking if these pages should use markdoc though?

scripts/transform-images.js
```diff
@@ -0,0 +1,37 @@
+import { existsSync } from 'fs';
```
TorstenDittmann, line 1: why do we need this, instead of re-using the vite optimizer?

TGlide: This is not for smaller image outputs, but rather, for the assets page, where there's the option to download the same image in multiple formats.

TorstenDittmann (approved):

PR #192, opened 2023-10-04
WIP: feat: add vale

loks0n:


gewenyu99 (commented):
There's a couple of things this can't do for sentence cases.

You already addressed the sentence case starting with numbers, but also proper nouns in sentence case is not enforced well.

PR #214, opened 2023-10-06
Blog posts API fix

TGlide:


TorstenDittmann (commented):
Why is this needed?

We actually want to prerender everything all the time 👍🏻

TGlide:
> Why is this needed?
> 
> We actually want to prerender everything all the time 👍🏻

This is an API endpoint route. The only reason I added the option to prerender on Vercel, is to not break the static adapter we're using there, but @PineappleIOnic needs to access it as an API in prod

TorstenDittmann (approved):

PR #224, opened 2023-10-10
Updated GitHub stars to latest

vermakhushboo:
## What does this PR do?

Updated number of GitHub stars to latest.


stnguyen90:
@TorstenDittmann, it might be good to use an env var for the star count since there are multiple places where it needs to be updated. Either that or make a component? Eventually, we should probably automate it.

TorstenDittmann (approved):

PR #227, opened 2023-10-11
QA fixes batch 3

TGlide:
## What does this PR do?

Several fixes


src/lib/components/Select.svelte
TorstenDittmann, line 1: https://github.com/appwrite/website/assets/1759475/35c7fb94-2ef5-4941-8bec-b82571ca8c88

kinda hard to spot that you can scroll, would it be possible to have the last item visible by its half?

src/lib/layouts/Main.svelte
```diff
@@ -109,6 +110,22 @@
 
         return $scrollInfo.deltaDirChange < 200;
     })();
+
+    let showDiscordBanner = false;
```
TorstenDittmann, line 114: can we make this more generic? it can potentially be any kind of banner

PR #261, opened 2023-10-18
Elad style fixes 1

elad2412:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)

### DIFF
--- a/src/scss/7-components/_main-header.scss
+++ b/src/scss/7-components/_main-header.scss
@@ -7,12 +7,14 @@
   --p-main-header-border-color: rgb(0,0,0, 0.1);
 
   position:sticky; z-index:999; inset-block-start:0; background-color:hsl(var(--p-body-bg-color));
-  border-block-end: solid 1px var(--p-main-header-border-color);
-  padding-inline: clamp(1.25rem, 4vw, 120rem); 
+  border-block-end: solid pxToRem(1) var(--p-main-header-border-color);
+  min-block-size:pxToRem(73);
   
   &.is-docs {
-    padding-inline: pxToRem(32);
-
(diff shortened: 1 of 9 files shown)


TGlide:
@elad2412 there's a merge conflict

src/scss/7-components/_main-header.scss
```diff
@@ -7,23 +7,22 @@
...
   position:sticky; z-index:999; inset-block-start:0; background-color:hsl(var(--p-body-bg-color));
-  border-block-end: solid 1px var(--p-main-header-border-color);
-  padding-inline: clamp(1.25rem, 4vw, 120rem); 
-  
-  &.is-docs {
-    padding-inline: pxToRem(32);
+  padding-inline:pxToRem(32); border-block-end: solid pxToRem(1) var(--p-main-header-border-color);
+  min-block-size:pxToRem(73);
 
+  &.is-special-padding {
```
TGlide, line 13: We only use the `main-header` in two places. One is `Main.svelte`, and the other is `Docs.svelte`. 

Instead of having one extra class per each, wouldn't it make more sense for one to use the default padding, and the other overrides it?

TGlide: Similar to what it was before actually

elad2412: @TGlide The particular padding for the homepages, the calculated dynamic, isn't so healthy for maintaining.

this is the reason why I separated it only while using special class.

TGlide: > isn't so healthy for maintaining
Why?

TGlide (approved):

PR #266, opened 2023-10-19
feat: new redirects

TorstenDittmann:


src/lib/components/FooterNav.svelte
```diff
@@ -46,7 +46,7 @@
                 },
                 {
                     label: 'Status',
-                    href: 'https://status.appwrite.io',
+                    href: 'https://www.appwrite.online',
```
TGlide, line 49: This leads to a page without a SSL certificate, and redirects to BetterStack's landing page

TGlide (approved):

PR #323, opened 2023-11-02
Fix FOUC

TGlide:
## What does this PR do?

Fixes FOUC caused by discord banner


src/routes/+layout.server.ts
TorstenDittmann, line 1: vercel needs prerendered routes 😆

PR #358, opened 2023-11-17
feat: add aria-labels to nav elements

### LINKED ISSUE #294: 📚 Documentation: multiple <nav> elements should be labelled [a11y][TheA11y100]
## Issue
There are multiple `<nav>` elements on each page.  When you have more than one `<nav>` you should add a label to each.

## Why it matters
When you have more than one `<nav>` element you should add a label to each to aid navigation with assistive tech and to differentiate them, otherwise when navigate via landmarks it can be confusing which navigation a user is targetting.

## Suggested fix
Add `aria-label`s to each of the `<nav>` elements to identify their purposes. 

For example:
Nav at the top could be `<nav aria-label="main">` and the nav at the side could be `<nav aria-label="docs">` (these could be better, if someone can think of more descriptive names for each navigation element. You do not need to say "navigation" as part of the text.)

## Relevant Guidance
[multiple navigation landmarks](https://www.aditus.io/patterns/multiple-navigation-landmarks/)

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

CBID2:
## What does this PR do?
This PR adds an aria-label to `<nav>` elements. This to make the site more accessible for users who use screen readers to navigate the site.  

## Test Plan

To test this, I plan on using an accessiblity chrome extension to see if it conforms to accessibility standards.

## Related PRs and Issues
Closes #294 

Yes


src/lib/components/FooterNav.svelte
```diff
@@ -78,7 +78,7 @@
         };
 </script>
 
-<nav
+<nav aria-label="Footer Navigation"
```
TGlide, line 81: Maybe just `Footer` is alright

src/lib/layouts/Docs.svelte
```diff
@@ -111,7 +111,7 @@
                         width="130"
                     />
                 </a>
-                <nav class="aw-main-header-nav">
+                <nav class="aw-main-header-nav" aria-label="Docs" >
```
TGlide, line 114: This is confusing, as it does not describe the purpose of this navigation itself. Maybe `Top` or `Main` would be better?

src/lib/layouts/Main.svelte
```diff
@@ -201,7 +201,7 @@
                         width="130"
                     />
                 </a>
-                <nav class="aw-main-header-nav">
+                <nav class="aw-main-header-nav" aria-label="github stars">
```
TGlide, line 204: Same here, maybe `Main` or top would be better

src/lib/layouts/Sidebar.svelte
```diff
@@ -33,7 +33,7 @@
 	}
 </script>
 
-<nav class="aw-side-nav">
+<nav class="aw-side-nav" aria-label="Side Navigation">
```
TGlide, line 36: We can omit the term `Navigation` here. A screen reader will read that it is a navigation element, and the aria label, so that would just be annoying to the user 😄

src/markdoc/tags/Tabs.svelte
```diff
@@ -26,7 +26,7 @@
 </script>
 
 <div class="aw-card is-normal u-margin-block-start-16" {...$root} use:root>
-	<nav class="tabs u-flex u-gap-16">
+	<nav class="tabs u-flex u-gap-16" aria-label="Tabs">
```
TGlide, line 29: I think we can remove the `nav` element here. Multiple tabs may be used in markdown, resulting in duplicated aria labels, which will be more frustrating then if these weren't navigation elements.

TGlide (changes requested):

TGlide:
@CBID2 Thank you for the PR!! I left a few comments 🙂

CBID2:
Hi @TGlide! :) I made the changes you suggested! :)

src/markdoc/tags/Tabs.svelte
```diff
@@ -26,7 +26,6 @@
 </script>
 
 <div class="aw-card is-normal u-margin-block-start-16" {...$root} use:root>
-	<nav class="tabs u-flex u-gap-16">
```
TGlide, line 29: Sorry, my wording was off here. By removing the nav element, I meant swapping it with something else, e.g. a div. Otherwise we lose the styling we had here.

CBID2:
Hi @TGlide! :) I made the changes! :)

TGlide (approved):
Awesome 😄

PR #382, opened 2023-11-23
Fix formatting & add endpoints to refereneces

TGlide:


src/lib/utils/specs.ts
```diff
@@ -2,310 +2,321 @@ import { OpenAPIV3 } from 'openapi-types';
...
+     * Exceptions for Android SDK.
+     */
+    const isAndroidJava =
+        platform === Platform.ClientAndroidJava || platform === Platform.ServerJava;
+    const isAndroidKotlin =
+        platform === Platform.ClientAndroidKotlin || platform === Platform.ServerKotlin;
+    const isAndroid = isAndroidJava || isAndroidKotlin;
+    const isAndroidServer = platform === Platform.ServerJava || platform === Platform.ServerKotlin;
+    const api = await getApi(version, platform);
+    console.log({ api });
```
TorstenDittmann, line 219: leftover

src/lib/utils/specs.ts
```diff
@@ -2,310 +2,321 @@ import { OpenAPIV3 } from 'openapi-types';
...
-							const schema = resolveReference(schemas as OpenAPIV3.ReferenceObject, api);
-							models.push({
-								id,
-								name: schema?.description ?? ''
-							});
-						}
-					}
-				}
+    for (const { method, value, url } of iterateAllMethods(api, service)) {
+        // console.log({ method, value });
```
TorstenDittmann, line 237: leftover

TorstenDittmann (approved):
Please make sure it goes through a design review 👍🏻

eldadfux:
Reminder to add some padding between the sticky top border and the `path` code block

eldadfux:
We're also missing the `/v1/` prefix for the paths.

gewenyu99 (approved):
Nice <3

Maybe address should be on the left tbh, let's talk about this with Chen later.

gewenyu99 (changes requested):
@TGlide Check Asana, there are some changes Chen asked for :)

gewenyu99:
image
One last comment, can we make the copy button here select only the path not the method?

Copy `/getAccount` vs `GET /getAccount` as a figurative example.

gewenyu99 (approved):
Other than my comment about copy, if that's difficult/not possible right now, we can merge this as is.

gewenyu99:
LGTM, paste is fixed

PR #430, opened 2023-12-19
- Header Docs

elad2412:
Add Discord Button, Include Side-nav for mobile.


TGlide:
I wonder if the button should be included in the mobile nav as well?

elad2412:
Out of date

Issue #187, opened 2023-10-03
🚀 Feature: Update Appwrite GitHub stars from GitHub API


OshriAsulin:
### 🔖 Feature description

In this issue, we aim to update the star count of the Sharkio GitHub repository using the GitHub API Instead of writing a static number.

### 🎤 Pitch

The "Retrieve Current Stars" feature should be implemented for the following reasons:

- Real-time Engagement: It provides users with up-to-the-minute information about the popularity of the Sharkio GitHub repository. This real-time data can encourage users to engage with the repository in various ways, such as starring it or exploring its contents.

- Transparency: Showing the current star count directly on the repository's interface promotes transparency. Users can quickly see how many others have found the repository valuable, which builds trust and credibility.

- User Convenience: Users no longer need to navigate to the GitHub website to check the star count. The feature offers a seamless and convenient way to access this information, enhancing the overall user experience.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

Sshivam26:
can i pls work on this as a part of event hacktoberfest!!

Traveller08:
@OshriAsulin Can I work on it?

stnguyen90:
We should automate pulling all the github stats on the [community](https://appwrite.io/community) page.

![image](https://github.com/appwrite/website/assets/1477010/c1a6d4b8-67e5-4ced-aea2-4f758d7fceca)

TGlide:
> We should automate pulling all the github stats on the [community](https://appwrite.io/community) page.
> 
[...]

PR #469, opened 2023-12-27
Feat: added github star from github api

jeet1desai:
## What does this PR do?

This PR gets all the current stars of appwrite repo from the github API

## Related PRs and Issues

#187 

Yes


TorstenDittmann:
I am afraid this will run into rate limits and should be done in build time.

Even if we implement this in build time, we are going to run into rate limits in local development for sure. For now it will probably make more sense to have a single constant for all places.

PR #490, opened 2024-01-04
Top Banner

elad2412:
- upgrade to new design
- added to mobile header as well


elad2412:
Was meant to be done for Console - not needed

PR #491, opened 2024-01-04
Discord threads

TGlide:


dcowan-london:
Would it not make more sense to have messages listed oldest first?

And also, username rendering (instead of Discord user IDs).

PR #514, opened 2024-01-11
init ticket page

TGlide:


.vscode/settings.json
TorstenDittmann, line 1: can we put .vscode into gitignore :)

package.json
```diff
@@ -56,16 +59,38 @@
     },
     "type": "module",
     "dependencies": {
+        "@appwrite.io/console": "^0.4.2",
         "@appwrite.io/pink": "0.1.0-next.9",
         "@appwrite.io/pink-icons": "0.1.0-next.9",
         "@appwrite.io/repo": "github:appwrite/appwrite#main",
+        "@resvg/resvg-js": "^2.6.0",
         "@splinetool/viewer": "0.9.455",
+        "appwrite": "^13.0.1",
```
TorstenDittmann, line 68: you should be able to just use `@appwrite.io/console` instead

scripts/download-contributor-data.js
```diff
@@ -0,0 +1,45 @@
...
+
+const outputFile = `./src/lib/contributors.ts`;
+
+async function fetchContributors() {
+    let page = 1;
+    let contributorsData = [];
+    let hasMoreData = true;
+
+    while (hasMoreData) {
+        console.log(page);
```
TorstenDittmann, line 14: is this useful?

src/lib/appwrite/init.ts
```diff
@@ -0,0 +1,12 @@
+import { PUBLIC_APPWRITE_PROJECT_INIT_ID } from '$env/static/public';
+import { Client, Account, Databases } from 'appwrite';
```
TorstenDittmann, line 2: here we can use the console sdk too

.vscode/settings.json
TGlide: Even the custom code snippets? They're useful 😄

scripts/download-contributor-data.js
TGlide: The log? Just to keep track of how progress is going.

TorstenDittmann: Can we make the log more meaningful instead of just numbers 😁

.vscode/settings.json
TorstenDittmann: some people have their own .vscode settings that would conflict 👍🏻

package.json
TorstenDittmann: this one 👀

PR #531, opened 2024-01-18
feat: announce assistant changes

loks0n:


eldadfux (commented):
i think this visual is for the old brand, let's get something new here

TorstenDittmann (approved):

PR #560, opened 2024-01-29
Update Apple Quick Start Docs

### LINKED ISSUE #559: 📚 Documentation: Apple Quick-Start Snippet Syntax Error
### 💭 Description

The quick start docs for Apple have a syntax error in the code.

In the [Create a login page](https://appwrite.io/docs/quick-starts/apple/#step-5) snippet, the following code is shown:

```swift
try await Appwrite.onRegister(
     viewModel.email,
     viewModel.password
)
```
`Error`
Instance member 'onRegister' cannot be used on type 'Appwrite'; did you mean to use a value of this type instead?

```swift
try! await Appwrite.onLogin(
     viewModel.email,
     viewModel.password
)
```
`Error`
Instance member 'onLogin' cannot be used on type 'Appwrite'; did you mean to use a value of this type instead?

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

Kilo-Loco:
## What does this PR do?

- Fixes #559 by creating an instance of an "Appwrite object" which can be referenced inside the `body` of the `ContentView`
- Removes references to OAuth since the current steps don't implement OAuth and it is not an ideal flow for a newcomer to Appwrite
- Updates the functionality of the final app to switch between screens based on session status.

 If this PR is accepted, similar changes need to be made to Android and Flutter to maintain parity between the mobile platform quick start pages. I believe these changes will shorten the path to a newcomer becoming familiar with Appwrite and provides a functioning app as a result to following the steps listed.

## Test Plan

Visit `docs/quick-starts/apple` in the project or in browser to see changes.

## Related PRs and Issues

More Issues and PRs will be created as a result of this PR being merged.

### Have you read the [Contributing Guidelines on issues]

Yes

### DIFF
--- a/src/routes/docs/quick-starts/apple/+page.markdoc
+++ b/src/routes/docs/quick-starts/apple/+page.markdoc
@@ -47,7 +47,18 @@ You can skip optional steps.
 
 {% /section %}
 
-{% section #step-3 step=3 title="Create Appwrite Object" %}
+{% section #step-3 step=3 title="Add the Appwrite SDK" %}
+To add the Appwrite SDK for Apple as a dependency, open the **File** menu and click **Add Packages**.
+
+In the **Package URL** search box, enter https://github.com/appwrite/sdk-for-apple. 
+
+Once the SDK is found, select **Up to Next Major Version** as your **Dependency Rule** and click **Add Package**.
+
+When dependency resolution is complete, click **Add Package** again to add the SDK package to your target.
+
+{% /section %}
+
+{% section #step-4 step=4 title="Create Appwrite Object" %}
 Find your project's ID in the **Settings** page. 
 
 {% only_dark %}
@@ -109,7 +120,8 @@ class Appwrite {
 }
 ```
 {% /section %}
-{% section #step-4 step=4 title="Create a login page" %}
+
+{% section #step-5 step=5 title="Create a login page" %}
 Add the following code to `ContentView.swift`.
 
 ```swift
@@ -164,6 +176,6 @@ struct ContentView: View {
 ```
 {% /section %}
 
-{% section #step-5 step=5 title="All set" %}
+{% section #step-6 step=6 title="All set" %}
 Run your project by clicking **Start active scheme** in Xcode.
 {% /section %}
--- a/src/routes/docs/quick-starts/apple/+page.markdoc
+++ b/src/routes/docs/quick-starts/apple/+page.markdoc
@@ -47,48 +47,7 @@ You can skip optional steps.
 
 {% /section %}
 
-{% section #step-3 step=3 title="Add the Appwrite SDK" %}
-To add the Appwrite SDK for Apple as a dependency, open the **File** menu and click **Add Packages**.
-
-In the **Package URL** search box, enter https://github.com/appwrite/sdk-for-apple. 
-
-Once the SDK is found, select **Up to Next Major Version** as your **Dependency Rule** and click **Add Package**.
-
-When dependency resolution is complete, click **Add Package** again to add the SDK package to your target.
-
-In order to allow creating OAuth sessions, the following URL scheme must be added to your **Info.plist** file.
-
-```xml
-<key>CFBundleURLTypes</key>
-<array>
-<dict>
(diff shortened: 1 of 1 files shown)


src/routes/docs/quick-starts/apple/+page.markdoc
```diff
@@ -56,39 +56,9 @@ Once the SDK is found, select **Up to Next Major Version** as your **Dependency
...
-```swift
-func scene(_ scene: UIScene, openURLContexts URLContexts: Set<UIOpenURLContext>) {
-    guard let url = URLContexts.first?.url,
-        url.absoluteString.contains("appwrite-callback") else {
-        return
-    }
-
-    WebAuthComponent.handleIncomingCookie(from: url)
-}
-```
```
abnegate, line 88: Let's keep this here

src/routes/docs/quick-starts/apple/+page.markdoc
```diff
@@ -137,64 +106,71 @@ class Appwrite {
...
-    @Published var password: String = ""
-}
-
 struct ContentView: View {
-    @ObservedObject var viewModel = ViewModel()
+    @State var email: String = ""
+    @State var password: String = ""
+    @State var session: Session?
+    
+    let appwrite = Appwrite()
```
abnegate, line 135: Let's keep the `ViewModel`, but move the `Appwrite` instance into it. We can also add proxy methods for `onRegister` and `onLogin` in the view model to call the appwrite functions

src/routes/docs/quick-starts/apple/+page.markdoc
Kilo-Loco: I thought it might be a good idea to remove this since it adds unnecessary steps to get the user up and running with   the `createEmailSession` flow. Wouldn't leaving the code there cause some confusion for newcomers to Appwrite?

src/routes/docs/quick-starts/apple/+page.markdoc
Kilo-Loco: I can go either way on `ViewModel` but I am recommending that we remove it since view models are a debatable topic in the SwiftUI community. I also feel like it adds unnecessary complexity to the docs and would be better suited for tutorials instead. What do you think?

src/routes/docs/quick-starts/apple/+page.markdoc
abnegate: I agree it's unnecessary if you only want to create an email/password flow, but I think it's important to have it here for users who do want to use OAuth; this information is otherwise a bit harder to find. What do you think @gewenyu99?

src/routes/docs/quick-starts/apple/+page.markdoc
abnegate: It's good practice to separate the data from the view in any framework, to promote modularity and improve testability. I agree it adds a bit of complexity but we need to strike a balance between simplicity and best practice. In this case, I think it's better to leave it. 

Although we could name it better like `LoginCredentials` or similar instead of just `ViewModel` to make it easier to reason about

src/routes/docs/quick-starts/apple/+page.markdoc
EVDOG4LIFE: Bumping - @gewenyu99

adityaoberai:
@Kilo-Loco I am closing this PR since there has been no activity here in over a year. If you would like to make any new changes, please [create an issue](https://github.com/appwrite/website/issues/new/choose) first and submit a PR once that issue is assigned to you.

For this PR, it may make a lot more sense to first discuss all opinions in an Issue before we jump to any changes since we have some differences of perspective, in case you'd like to open this up again.

src/routes/docs/quick-starts/apple/+page.markdoc
adityaoberai: Agreeing with @abnegate here on retaining ViewModel logic

PR #720, opened 2024-02-28
doc: update angular quickstart, fix function call in code example

ddiebel:
## What does this PR do?

The function call in the example code was outdated and now it works with the latest client.

## Test Plan

build the angular quick-start

## Related PRs and Issues
-

Yes


gewenyu99:
It's actually the other way around. The latest client use createEmailPasswordSession, but it's not supported on Cloud. It's the first time we're dealing with Cloud and self-hosted version miss match, and we're going to add some long term solutions (versioning) to fix this.

Closing this for now, thanks for the help!

PR #786, opened 2024-03-11
Long links break into new line

elad2412:
Fix the issue. If we will have specific locations where links will need inline-flex behavior will support different solution to them.

(Issues happened on big chunl of text on Docs)


stnguyen90 (approved):
wrapping problem is fixed

image


I also poked around other pages and things seem okay...

PR #919, opened 2024-04-25
feat: terms regarding unsolicited sms

loks0n:


MayEnder:
Looks great but should we maybe loop in Legal?

TorstenDittmann (approved):

PR #977, opened 2024-05-20
Performance updates

elad2412:
- Replace images with smaller size
-  Update lazyload for several images


static/images/animations/tech-dark.png
stnguyen90, line 1: We use this plugin to automatically optimize images on build so I don't know if this is necessary:

https://github.com/appwrite/website/blob/468c5463a01a8b9cf49d4735e20b7765500e3f23/vite.config.ts#L18-L20

For example, this file's size on prod:

image

@TorstenDittmann?

TorstenDittmann (approved):

Issue #974, opened 2024-05-17
🚀 Enhancement: adding rss feed to blog


LimaneGaya:
### 🔖 Enhancement description

An rss feed to the blog section of the website

### 🎤 Pitch

it's a good way to keep people updated on the latest and greatest of Appwrite, it's a common way every other providers use to keep interested users updated.

### 👀 Have you spent some time to check if this issue has been raised before?

- [X] I checked and didn't find similar issue

### 🏢 Have you read the Code of Conduct?

- [X] I have read the [Code of Conduct](https://github.com/appwrite/.github/blob/main/CODE_OF_CONDUCT.md)

ItzNotABug:
Doesn't one exist already?
https://appwrite.io/blog/rss.xml

LimaneGaya:
Yes but i seems to not show in correct format i can see in the code and in #337 that it should work, but my [rss reader](https://feeder.co) does not find it and going to https://appwrite.io/blog/rss.xml show up like follows:

<details closed><summary>appwrite.io/blog/rss.xml</summary>

```xml
[...]

PR #1035, opened 2024-06-14
Fix: Broken `XML` due to Title and Urls in `RSS`

ItzNotABug:
## What does this PR do?

An `&` and a ` ` (space) broke the xml format. This should fix the issue.

## Test Plan

Manual check via `https://validator.w3.org/feed`.

## Related PRs and Issues

* #974.

Yes.

### DIFF
--- a/src/routes/blog/rss.xml/+server.ts
+++ b/src/routes/blog/rss.xml/+server.ts
@@ -3,7 +3,7 @@ import { posts } from '../content';
 
 export const prerender = true;
 
-function encodeTitle(str: string): string {
+function encodeText(str: string): string {
     return str.replace(/&/g, '&amp;')
         .replace(/</g, '&lt;')
         .replace(/>/g, '&gt;')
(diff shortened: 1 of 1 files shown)


src/routes/blog/rss.xml/+server.ts
```diff
@@ -3,23 +3,40 @@ import { posts } from '../content';
...
     <description>Appwrite is an open-source platform for building applications at any scale, using your preferred programming languages and tools.</description>
     ${posts.map((post) => `<item>
-        <title>${post.title}</title>
+        <title>${encodeTitle(post.title)}</title>
         <pubDate>${post.date.toUTCString()}</pubDate>
-        <link>https://appwrite.io${post.href}</link>
-        <guid>https://appwrite.io${post.href}</guid>
+        <link>${encodeUrl(post.href)}</link>
+        <guid>${encodeUrl(post.href)}</guid>
         <description>${post.description}</description>
```
TorstenDittmann, line 31: I tihnk descrition needs to be encoded too - I am not sure about the urls though

ItzNotABug: I could add encoding to description, kinda makes sense.
One of the urls had a space and that would really break the xml, so I added encoding to urls too.

ItzNotABug:
@TorstenDittmann should we keep description if `null`?

TorstenDittmann:
> @TorstenDittmann should we keep `description` if `null`?

There should always be a description 👍🏻 

can we check and use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI for the urls?

ItzNotABug:
> > @TorstenDittmann should we keep `description` if `null`?
> 
> There should always be a description 👍🏻
> 
> can we check and use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI for the urls?

I am actually using `encodeURI`, just in a [function](https://github.com/appwrite/website/blob/07f5c77083c3fc35086e723f110398a533eedea9/src/routes/blog/rss.xml/%2Bserver.ts#L16C1-L18C2). Or am I missing something here 🤔?

TorstenDittmann (approved):

PR #1041, opened 2024-06-17
Integrations pages

elad2412:
- 3 new Integration pages


src/lib/components/embla/Embla.svelte
```diff
@@ -0,0 +1,168 @@
...
+<style>
+    .embla {
+        overflow: hidden;
+        position: relative;
+    }
+
+    .embla__container {
+        display: flex;
+    }
+</style>
```
TorstenDittmann, line 168: feel free to use sass here, you also dont need to care about css scopes with svelte and can just use `.container` 👍🏻

src/lib/components/embla/EmblaSlide.svelte
```diff
@@ -0,0 +1,19 @@
+<li class="embla__slide web-carousel-item">
+    <div class="embla__slide__number">
+        <slot />
+    </div>
+</li>
+
+<style lang="scss">
+    @use '$scss/abstract' as *;
+
+    .embla__slide {
```
TorstenDittmann, line 10: same as here, with svelte you can just do `.slide` or even `div` or `li`

src/lib/components/embla/Embla.svelte
TorstenDittmann, line 1: what do you think about naming the component library agnostic, like `carousel` over `embla`

src/lib/components/embla/index.ts
```diff
@@ -0,0 +1,7 @@
+import EmblaRoot from './Embla.svelte';
+import EmblaSlide from './EmblaSlide.svelte';
+
+export const Embla = Object.assign({
+    Root: EmblaRoot,
+    Slide: EmblaSlide
+});
```
TorstenDittmann, line 7:
```suggestion
export { default as Root } from './Embla.svelte';
export { default as Slide } from './EmblaSlide.svelte';
```

does this work? 👀

src/lib/integrations/Carousel.svelte
```diff
@@ -0,0 +1,69 @@
+<script lang="ts">
+    let carousel: HTMLElement;
+
+    let gap = 32;
```
TorstenDittmann, line 4: this can be a `const` right?

src/lib/integrations/index.ts
```diff
@@ -0,0 +1,14 @@
...
+
+export type ResultType<T> = FuseResult<T>[];
+
+import CarouselRoot from './Carousel.svelte';
+import CarouselSlide from './CarouselSlide.svelte';
+
+export const Carousel = Object.assign({
+    Root: CarouselRoot,
+    Slide: CarouselSlide
+});
```
TorstenDittmann, line 14: if the previously mentioned approach works - we can do it here too 👍🏻

src/lib/integrations/SvelteFuse.svelte
```diff
@@ -0,0 +1,17 @@
+<script lang="ts" generics="T">
```
TorstenDittmann, line 1: we might wanna investigate to use server endpoints for searching - might scale better in the long run 👍🏻

src/scss/2-resets/_typography.scss
```diff
@@ -1,4 +1,7 @@
 @use '../abstract' as *;
+html {
+	scroll-behavior: smooth;
+}
```
TorstenDittmann, line 4: this will affect all pages, not sure if its intended

adityaoberai:
@TorstenDittmann the page we see under `routes/integrations/article`, can we convert it into a markdoc layout like posts in the blog?

adityaoberai:
> > @TorstenDittmann @thejessewinton  the page we see under `routes/integrations/article`, can we convert it into a markdoc layout like posts in the blog?
> 
> Definitely; on my end it will likely be blocked by Init work, so how soon do you need it?

That's a good point. Gonna take this discussion to Discord.

src/lib/components/embla/Embla.svelte
thejessewinton: This is actually a requirement so `embla` knows which element to attach to.

https://www.embla-carousel.com/get-started/svelte/

src/lib/components/embla/Embla.svelte
thejessewinton: Great, will do!

src/lib/components/embla/EmblaSlide.svelte
thejessewinton: 👍🏼

src/lib/components/embla/index.ts
thejessewinton: It does, just requires multiple imports. Easy change!

src/lib/integrations/SvelteFuse.svelte
thejessewinton: Was already thinking about this; will throw this in as a busy work task between Init stuff this week.

ArmanNik (approved):

PR #1074, opened 2024-06-26
feat: sentry

TorstenDittmann:
## What does this PR do?

- adds sentry to prod and stage
- uploads sourcemaps

✅


src/hooks.client.ts
```diff
@@ -0,0 +1,21 @@
+import { handleErrorWithSentry, replayIntegration } from "@sentry/sveltekit";
+import * as Sentry from '@sentry/sveltekit';
+
+Sentry.init({
+  dsn: 'https://27d41dc8bb67b596f137924ab8599e59@o1063647.ingest.us.sentry.io/4507497727000576',
```
Meldiron, line 5: Due to this, both Stage and Production will go to same project. We might want another ENV variable for DSN

src/hooks.server.ts
```diff
@@ -1,8 +1,14 @@
+import * as Sentry from '@sentry/sveltekit';
 import type { Handle } from '@sveltejs/kit';
 import redirects from './redirects.json';
 import { sequence } from '@sveltejs/kit/hooks';
 import { BANNER_KEY } from '$lib/constants';
 
+Sentry.init({
+    dsn: "https://27d41dc8bb67b596f137924ab8599e59@o1063647.ingest.us.sentry.io/4507497727000576",
```
Meldiron, line 8: Same here

src/hooks.client.ts
TorstenDittmann: should be fine, we can make it dependend on the url 👍🏻

src/hooks.server.ts
Meldiron: If we dont do it as configurable variable from ENV (GitHub secrets), we should make it const and re-use, so we can update it in 1 place in future

src/routes/sentry-example/+page.svelte
```diff
@@ -0,0 +1,97 @@
+<!-- 
+This is just a very simple page with a button to throw an example error.
+Feel free to delete this file and the entire sentry route.
```
Meldiron, line 3: If we plan to keep it, we can remove this line of the comment

src/routes/sentry-example/+server.js
```diff
@@ -0,0 +1,6 @@
+// This is just a very simple API route that throws an example error.
+// Feel free to delete this file and the entire sentry route.
```
Meldiron, line 2: Same here, no need for this line

src/routes/sentry-example/+page.svelte
TorstenDittmann: just wanna keep it for testing stage 👍🏻

Meldiron (approved):

PR #1129, opened 2024-07-17
Fix `TOC` on Desktop & Mobile

ItzNotABug:
## What does this PR do?

As per the discussions, TOC should not show on desktops. Also fixes broken on mobile.

## Test Plan

Manual.

## Related PRs and Issues

N/A.

Yes.


src/routes/assets/+page.svelte
```diff
@@ -64,22 +66,27 @@
...
+                                <span class="icon-menu-alt-4" aria-hidden="true" />
+                            </span>
+                        </button>
+                    </div>
                 </header>
-                <TocNav />
+                <TocNav bind:showToc />
                 <main class="web-grid-120-1fr-auto-main /web-is-mobile-closed" id="main">
-                    <div class="web-content">
+                    <div class="web-content" class:web-is-mobile-closed={showToc && !showToc}>
```
loks0n, line 89: `{showToc && !showToc}` should always evaluate false right?

ItzNotABug: I think we might not even need this class here. This PR still requires further fixes anyways.

src/lib/components/TocNav.svelte
```diff
@@ -12,94 +13,194 @@
...
+            easing,
+            css: (t: number) => {
+                return `
+                    opacity: ${t};
+                    height: ${t * initialHeight}px;
+                    overflow: hidden;
+                `;
+            }
+        };
+    }
```
ernstmul, line 39: Can we also do this animation with CSS? I think that would better fit the direction we're moving in. Wdyt @thejessewinton?

src/markdoc/layouts/Policy.svelte
```diff
@@ -107,7 +98,20 @@
         opacity: 0;
     }
 
-    .toc-btn {
-        transition: translate 0.3s ease;
+    @media (max-width: 768px) {
+        .container {
+            padding-left: 0 !important;
+            padding-right: 0 !important;
```
ernstmul, line 104: Are these `!important`'s really needed?

src/routes/assets/+page.svelte
```diff
@@ -628,4 +622,21 @@
             }
         }
     }
+
+    @media (max-width: 768px) {
+        .container {
+            padding-left: 0 !important;
+            padding-right: 0 !important;
```
ernstmul, line 629: Same question about important

src/lib/components/TocNav.svelte
ItzNotABug: Not sure if we probably can (due to `t * initialHeight`, but correct me if I am wrong), but I think I've seen a similar use-case somewhere around the repo. Lmk what we decide.

src/markdoc/layouts/Policy.svelte
ItzNotABug: I tested again, not needed. removed.

src/routes/assets/+page.svelte
ItzNotABug: Addressed.

src/lib/components/TocNav.svelte
ItzNotABug: ping @ernstmul @thejessewinton.

ernstmul (approved):

PR #1082, opened 2024-07-01
Docs: Binary & Stream Executions

choir27:
## What does this PR do?

- [x] Update Develop > response
    - [x] Break down different sections for each request/response type
    - [x] Replace req.bodyRaw with req.bodyText
    - [x] Don’t use req.body because it may not detect so replace with bodyText or bodyJson depending on expected input
    - [x] Add deprecation notice for req.bodyRaw and req.bodyText and res.send
    - [x] Res.send should be res.text

## Test Plan

X

## Related PRs and Issues

X

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes

PR #1140, opened 2024-07-19
Docs: Binary & Stream Executions

choir27:
## What does this PR do?

- [x] Update Develop > response
    - [x] Break down different sections for each request/response type
    - [x] Replace req.bodyRaw with req.bodyText
    - [x] Don’t use req.body because it may not detect so replace with bodyText or bodyJson depending on expected input
    - [x] Add deprecation notice for req.bodyRaw and req.bodyText and res.send
    - [x] Res.send should be res.text
    
Look at https://github.com/appwrite/website/pull/1082 to view comments/feedback for previous PR

## Test Plan

Manually tested a function on the preview instance using `createExecution` to view the response.

## Related PRs and Issues

Manually created an Appwrite function and executed it in the preview instance to test the response of res.text, req.bodyJson, and req.bodyText.

Yes


src/routes/docs/products/functions/develop/+page.markdoc
```diff
@@ -465,11 +469,31 @@ If you pass data into an Appwrite Function, it'll be found in the request object
...
+
+{% table %}
+* Request
+* Description
+---
+* `req.bodyRaw`
+* Returns the raw body text data.
+---
+* `req.bodyText`
+* Returns binary data that was converted into text.
```
stnguyen90, line 482: This is a little misleading because it says "returns binary data", but what does it return?

src/routes/docs/products/functions/develop/+page.markdoc
```diff
@@ -694,6 +719,32 @@ If you need to send a response to the invoker of the function, such as a user, c
...
 There are several possible ways to send a response, explore them in the following Appwrite Function.
 
+### Response Types {% #response-types %}
+
+{% table %}
+* Response
+* Description
+---
+* `send`
+* Sends a plain text body as a response.
```
stnguyen90, line 729: send is deprecated, right? Probably not good to mention it as one that can be used.

stnguyen90 (changes requested):

PR #1162, opened 2024-07-29
Added feature that the logged in user is displayed with logout button…

shivanshukarn:
# Implement Persisted User Login State on Page Refresh

As per the code in docs, after logging in, if we refresh the page then the loggedInUser state is set to null as default and it will be changed only when the login function is called, but the user is still logged in. So I have integrated a feature that if the user is logged in then the logged-in component will be displayed.

## What does this PR do?

This PR implements a mechanism to persist the user's login state across page refreshes. Previously, the loggedInUser state was set to null on page reload and only updated when the login function was called, even though the user was still logged in. With this change, if the user is logged in, the logged-in component will be displayed immediately after a page refresh.

## Test Plan

To verify the changes:
1. Log in to the application.
2. Refresh the page.
3. Confirm that the logged-in component is displayed immediately after the refresh, indicating that the user's login state has been persisted.

## Related PRs and Issues

Yes, I have read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md).

### DIFF
--- a/src/routes/docs/quick-starts/nextjs/+page.markdoc
+++ b/src/routes/docs/quick-starts/nextjs/+page.markdoc
@@ -76,7 +76,7 @@ Create or update `app/page.js` file and add the following code to it.
 
 ```js
 "use client";
-import { useState } from "react";
+import { useState, useEffect } from "react";
 import { account, ID } from "./appwrite";
 
 const LoginPage = () => {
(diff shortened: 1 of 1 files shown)


shivanshukarn:
Implement Persisted User Login State on Page Refresh

PR #1172, opened 2024-08-05
Improve Title Context

ItzNotABug:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)


src/lib/layouts/DocsTutorial.svelte
```diff
@@ -3,77 +3,103 @@
...
 
-    export let title: string;
     export let toc: Array<TocItem>;
     export let currentStep: number;
-    export let back: string;
     export let date: string;
 
     export let tutorials: Array<Tutorial>;
 
+    const firstStepItem = tutorials[0];
```
ernstmul, line 15: If tutorials is an empty array, this would yield a nullpointer right?

src/lib/layouts/DocsTutorial.svelte
```diff
@@ -3,77 +3,103 @@
...
+    // currentStep starts from 1, the arrays start from 0.
+    const currentStepItem = (tutorials[currentStep - 1] ?? firstStepItem);
+
     $: nextStep = tutorials.find((tutorial) => tutorial.step === currentStep + 1);
     $: prevStep = tutorials.find((tutorial) => tutorial.step === currentStep - 1);
+
+    // `any` for compatibility with reactive variables.
+    function getCorrectTitle(tutorial: any | Tutorial, checkAt: number): string {
```
ernstmul, line 23: But it's always a Tutorial right? I would prefer to keep this one strict, and the cast the variable that is thrown into the function when needed.

ItzNotABug: Aah yes, it is. I think I added `any` because `prevStep` returns `any|Tutorial`, not sure why. The type should be available there.

src/lib/layouts/DocsTutorial.svelte
ItzNotABug: Possibly yes, although I doubt it'll be empty as there's each for a framework, split into different steps. Wdyt?

ernstmul: I would prefer to check it, just so we're sure that if something goes wrong upstream, it doesn't crash the site.

ernstmul (approved):

PR #1152, opened 2024-07-25
Feat: Add Pagination to Blog

ItzNotABug:
## What does this PR do?

Pagination on Blog Pages.

## Test Plan

Manual.

## Related PRs and Issues

* https://github.com/appwrite/website/issues/1038

### Have you read the [Contributing Guidelines on issues](https://github.com/appwrite/appwrite/blob/master/CONTRIBUTING.md)?

Yes.

PR #1226, opened 2024-08-17
Feat: Blog x Search

ItzNotABug:
## What does this PR do?

Parent PR -
* https://github.com/appwrite/website/pull/1152

## Test Plan

## Related PRs and Issues

(Write your answer here.)


TorstenDittmann (commented):
Lets make some changes to how the search/filter state is handled, lets make the category and search part of the query param and do the data fetching on the `+page.ts` 👍🏻

src/routes/blog/[[page]]/+page.svelte
```diff
@@ -34,6 +41,51 @@
         });
     });
 
+    let selectedCategory = $page.url.searchParams.get('category') ?? 'Latest';
+
+    const handleSearch = async (value: string) => {
```
Comment (ernstmul) [inline src/routes/blog/[[page]]/+page.svelte:46]: The argument here is always the value from the global query right? Why not make this function without an argument, and use the global `query`? Line 47 would then become something like `const queryLowercase = query.toLowerCase()`

src/routes/blog/[[page]]/+page.svelte
```diff
@@ -316,4 +468,113 @@
...
+    }
+
+    .is-selected {
+        color: hsl(var(--web-color-black));
+        background: hsl(var(--web-color-white));
+    }
+
+    .overflow-x-auto {
+        scrollbar-width: none;
+        -ms-overflow-style: none;
```
Comment (ernstmul) [inline src/routes/blog/[[page]]/+page.svelte:483]: I'm not too sure about our support levels, but don't think this is needed anymore right? Current Edge should pick

src/routes/blog/[[page]]/+page.svelte
```diff
@@ -316,4 +468,113 @@
...
+            max-width: 62.5%;
+        }
+    }
+
+    @media (max-width: 768px) {
+        .search-and-categories {
+            flex-wrap: wrap;
+            display: flex !important;
+            padding-block-start: 2rem;
+            flex-direction: column-reverse !important;
```
Comment (ernstmul) [inline src/routes/blog/[[page]]/+page.svelte:565]: Are these 2 importants really needed? Can you get rid of them by making the selector more specific?

src/routes/blog/[[page]]/+page.svelte
```diff
@@ -316,4 +468,113 @@
...
+        .categories-wrapper {
+            max-width: 90vw;
+        }
+
+        .search-and-categories > .mobile-search-bar {
+            align-self: baseline;
+        }
+
+        .search-and-categories > .web-input-text-search-wrapper {
+            min-inline-size: unset !important;
```
Comment (ernstmul) [inline src/routes/blog/[[page]]/+page.svelte:577]: Same !important question as above

src/routes/blog/[[page]]/+page.svelte
Comment (ItzNotABug) [reply src/routes/blog/[[page]]/+page.svelte:46]: ohh yeah that makes sense.

src/routes/blog/[[page]]/+page.svelte
Comment (ItzNotABug) [reply src/routes/blog/[[page]]/+page.svelte:483]: I wasn't sure too either, added it just for safety. wouldn't want a broken view on some older edge maybe. wdyt? should I remove - `ms-*`?

src/routes/blog/[[page]]/+page.svelte
Comment (ItzNotABug) [reply src/routes/blog/[[page]]/+page.svelte:565]: yes, we can remove those.

src/routes/blog/[[page]]/+page.svelte
Comment (ItzNotABug) [reply src/routes/blog/[[page]]/+page.svelte:577]: I think we can remove this too.

src/routes/blog/[[page]]/+page.svelte
Comment (ernstmul) [reply src/routes/blog/[[page]]/+page.svelte:483]: I double checked with Torsten, we can remove those!

src/routes/blog/[[page]]/+page.svelte
```diff
@@ -34,6 +41,59 @@
...
+
+    const handleSearch = async () => {
+        const searchQuery = query.toLowerCase();
+
+        const url = new URL($page.url);
+        /**
+         * Navigate to the first page on search/filter to ensure consistent
+         * navigation experience when changing categories or search queries.
+         */
+        url.pathname = '/blog';
```
Comment (TorstenDittmann) [inline src/routes/blog/[[page]]/+page.svelte:54]:
```suggestion
        /**
         * Navigate to the first page on search/filter to ensure consistent
         * navigation experience when changing categories or search queries.
         */
        const url = new URL('/blog', $page.url);
```

src/routes/blog/[[page]]/+page.svelte
```diff
@@ -246,28 +386,28 @@
...
-                    <FooterNav />
-                    <MainFooter />
-                </div>
+        </div>
+        <div>
+            <div class="container">
+                <FooterNav />
+                <MainFooter />
             </div>
         </div>
```
Comment (TorstenDittmann) [inline src/routes/blog/[[page]]/+page.svelte:412]: do we need the `div` around the `div.container`?

src/routes/blog/[[page]]/+page.svelte
```diff
@@ -246,28 +386,28 @@
                             {#if data.currentPage < data.totalPages}
                                 <a
                                     data-sveltekit-noscroll
-                                    class="flex navigation-button"
+                                    class="navigation-button flex"
                                     href="/blog/{data.currentPage + 1}"
```
Comment (TorstenDittmann) [inline src/routes/blog/[[page]]/+page.svelte:390]: the pagination will remove `category` or `search` queries 👍🏻 we need to add those here.

feel free to create a helper function that builds a link

src/routes/blog/content.ts
```diff
@@ -105,9 +106,16 @@ export const categories = Object.values(categoriesGlob).map((categoryList) => {
     };
 });
 
-export const getAllBlogEntriesWithAuthors = () => {
+export const normalizeCategory = (str: string) => str?.replace(/\s+/g, '-').toLowerCase();
```
TorstenDittmann, line 109: this may be a bit sketchy but good for now, we need a proper category map at some point I think :-)

ItzNotABug: we definitely do, there's quite inconsistency for categories mentioned and used across posts.

src/routes/blog/[[page]]/+page.svelte
Comment (ItzNotABug) [reply src/routes/blog/[[page]]/+page.svelte:412]: yes, for the footer's container. else the footer container just squeezes in width.

src/routes/blog/[[page]]/+page.svelte
Comment (ItzNotABug) [reply src/routes/blog/[[page]]/+page.svelte]: Done.

ernstmul (approved):

PR #1228, opened 2024-08-18
Fix sections numberings

ItzNotABug:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)


pnpm-lock.yaml
stnguyen90, line 1: Did this file need to be updated?

src/lib/layouts/DocsArticle.svelte
```diff
@@ -117,4 +117,13 @@
...
     </article>
-</main>
\ No newline at end of file
+</main>
+
+<style>
+    @media (min-width: 1280px) and (max-width: 1330px) {
+        .web-article-content {
+            /* original/default is 41.5rem */
+            max-inline-size: 40.5rem !important;
```
stnguyen90, line 126: This seems like it causes other content to be misaligned. For example, [Auth docs](http://1228.gkswsco.37.27.92.31.sslip.io/docs/products/auth):
image

stnguyen90 (changes requested):

src/lib/layouts/DocsArticle.svelte
ItzNotABug: Done.

pnpm-lock.yaml
ItzNotABug: Not sure, ig this was somehow just committed. I'll remove this file.

ItzNotABug: accidentally deleted, will restore and remove from this pr.

ItzNotABug: pheww, done!

src/lib/layouts/DocsArticle.svelte
```diff
@@ -16,11 +16,14 @@
...
 
     import { Feedback } from '$lib/components';
+    import { page } from '$app/stores';
 
     export let title: string;
     export let toc: Array<TocItem>;
     export let back: string | undefined = undefined;
     export let date: string | undefined = undefined;
+
+    $: isQuickStartArticle = $page.route.id?.includes('/docs/quick-starts/');
```
stnguyen90, line 26: This seems quite hacky since it doesn't catch pages that aren't quickstarts. For example, [migrations](https://appwrite.io/docs/advanced/migrations/firebase):

image

or [AI](https://appwrite.io/docs/products/ai/integrations/perplexity):

image


I think we need a better solution to handle all cases where we're using this numbered step.

stnguyen90 (dismissed):

src/lib/layouts/DocsArticle.svelte
ItzNotABug: Makes sense. I'll improve the logic for this. Although we could have a completely different component with content with sections but that feels like an overkill and the docs [.md] might require a good amount of refactors.

ItzNotABug: Done. It now only resizes if section with numbering is used on content. Let me know if this works now.

src/lib/layouts/DocsArticle.svelte
```diff
@@ -16,11 +16,22 @@
...
+
+    let reducedArticleSize = false;
+
+    onMount(() => {
+        // reduces size for articles using numbered sections.
+        const slotWrapper = document.querySelector('.slot-wrapper');
+        reducedArticleSize = !!slotWrapper?.querySelector(
+            '.web-article-content-section.is-with-line .web-numeric-badge'
+        );
+    })
```
stnguyen90, line 34: This still feels hacky. And what if they resize the window?

@TorstenDittmann, any ideas on how to best handle this?

ItzNotABug: @stnguyen90 Works on resize. The issue only occurs at certain breakpoints which the css handles.

ernstmul: But does it also work when I manually resize the window without a reload @ItzNotABug?

ItzNotABug: > But does it also work when I manually resize the window without a reload @ItzNotABug?

Just verified on a tutorial docs page, seems to work.

src/lib/layouts/DocsArticle.svelte
```diff
@@ -16,11 +16,22 @@
...
     export let title: string;
     export let toc: Array<TocItem>;
     export let back: string | undefined = undefined;
     export let date: string | undefined = undefined;
+
+    let reducedArticleSize = false;
+
+    onMount(() => {
+        // reduces size for articles using numbered sections.
+        const slotWrapper = document.querySelector('.slot-wrapper');
```
TorstenDittmann, line 30: can we do a svelte binding here with `bind:this` instead?

src/lib/layouts/DocsArticle.svelte
```diff
@@ -110,3 +127,12 @@
...
     </article>
 </main>
+
+<style>
+    @media (min-width: 1280px) and (max-width: 1330px) {
+        .reduced-article-size {
+            /* original/default is 41.5rem */
+            max-inline-size: 40.5rem !important;
+        }
+    }
```
TorstenDittmann, line 137: can we check and see to add this to the scss files if possible?

src/lib/layouts/DocsArticle.svelte
```diff
@@ -16,11 +16,22 @@
...
     export let date: string | undefined = undefined;
+
+    let reducedArticleSize = false;
+
+    onMount(() => {
+        // reduces size for articles using numbered sections.
+        const slotWrapper = document.querySelector('.slot-wrapper');
+        reducedArticleSize = !!slotWrapper?.querySelector(
+            '.web-article-content-section.is-with-line .web-numeric-badge'
+        );
```
TorstenDittmann, line 33: can we solve that with svelte context? this doesn't really age well with targetting classnames imo

ItzNotABug: Done.

src/lib/layouts/DocsArticle.svelte
```diff
@@ -13,14 +13,18 @@
...
+    import { scrollToTop } from '$lib/actions/scrollToTop';
 
     export let title: string;
     export let toc: Array<TocItem>;
     export let back: string | undefined = undefined;
     export let date: string | undefined = undefined;
+
+    let reducedArticleSize = false;
+
+    setContext('articleHasNumericBadge', () => (reducedArticleSize = true));
```
TorstenDittmann, line 27: you can do something like this to make it easier:

```suggestion
    const reducedArticleSize = setContext('articleHasNumericBadge', writable(false));
```

and then on the section components you can do:

```js
if (hasContext('articleHasNumericBadge')) getContext('articleHasNumericBadge').set(true);
```

src/scss/6-elements/text-tokens/_article-size.scss
```diff
@@ -0,0 +1,8 @@
+@use '../../abstract' as *;
+
+@media (min-width: 1280px) and (max-width: 1330px) {
+  .#{$p}-reduced-article-size {
+    /* original/default is 41.5rem */
+    max-inline-size: 40.5rem !important;
+  }
```
TorstenDittmann, line 7: can we put this where `  .#{$p}-reduced-article-size {` is defined in the first place, I dont think we need a new file for it 👍🏻

TorstenDittmann (approved):

PR #1241, opened 2024-08-20
Docs: Dynamic API keys

choir27:
## What does this PR do?

add content about dynamic key injected as APPWRITE_FUNCTION_API_KEY and accessing it during build

## Test Plan

Manually tested against preview instance

## Related PRs and Issues

X

Yes

### DIFF
--- a/src/routes/docs/products/functions/develop/+page.markdoc
+++ b/src/routes/docs/products/functions/develop/+page.markdoc
@@ -1350,6 +1350,8 @@ Dynamic API keys are the same as [API keys](/docs/advanced/platform/api-keys) bu
 They are generated in your functions per execution.
 However, you can only use dynamic API keys inside Appwrite functions.
 
+During the build process, dynamic API keys are injected as the environment variable `APPWRITE_FUNCTION_API_KEY`. Initialize `APPWRITE_FUNCTION_API_KEY` by navigating to your function > **Settings** > Environment variables. This environment variable doesn't need to be initialized in your function code or your `.env` file.
+
 Dynamic API keys grant access and operate without sessions.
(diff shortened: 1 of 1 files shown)


src/routes/docs/products/functions/develop/+page.markdoc
```diff
@@ -1350,6 +1350,8 @@ Dynamic API keys are the same as [API keys](/docs/advanced/platform/api-keys) bu
 They are generated in your functions per execution.
 However, you can only use dynamic API keys inside Appwrite functions.
 
+During the build process, dynamic API keys are injected as the environment variable `APPWRITE_FUNCTION_API_KEY`. Initialize `APPWRITE_FUNCTION_API_KEY` by navigating to your function > **Settings** > Environment variables. This environment variable doesn't need to be initialized in your function code or your `.env` file.
```
loks0n, line 1353: I feel like 'injected' is a complicated word that has other implications.

Is it possible to reword this with simpler language?

loks0n (changes requested):

src/routes/docs/products/functions/develop/+page.markdoc
Meldiron: Agree, we can say `automatically provided` instead of `inject`.

Also part about adding it to environment variables in function settings is not right. They don't need to do anything, simply use it during build step. We will add it for them. their env variables card in function settings can be empty, and they will still have this env variable. just like ENDPOINt or PROJECT_ID ones, and others.

Meldiron (approved):

stnguyen90 (approved):

PR #1269, opened 2024-08-23
Docs: Update local develop

choir27:
## What does this PR do?

Local development add the following section:
    --port: functions run on port 3000 on default, or closest available, and with this parameter you can specify port you want
    --function-id: lets you autoselect function, so you don't need to click through the list every time. Cool for shortcuts like npm run api
    
Local development section about hot reloading
- what is hot reloading
-how it works
- complied languages always re-build, but interpreted only when you change dependency file, otherwise it's a quick file swap. We could also have list of file names for each runtime that are considered dependency files, like package.json or go.mod

## Test Plan

Test parameters

## Related PRs and Issues

X

Yes


src/routes/docs/products/functions/develop-locally/+page.markdoc
```diff
@@ -71,6 +75,81 @@ You can use headers like dynamic API keys in your function, which give you acces
 Learn more about dynamic API keys
 {% /arrow_link %}
 
+# Hot reloading {% #hot-reloading %}
+
+You can hot reload your locally developed functions, which means you can update the function code and see the changes happen in real time without restarting. Compiled languages must translate the source code into machine code, causing the system to rebuild continuously. However, interpreted languages only rebuild when you change the dependency file. Otherwise, updates can be made through a file swap.
```
loks0n, line 80: `Otherwise, updates can be made through a file swap.`

What does this mean? I'm not sure I understand it

src/routes/docs/products/functions/develop-locally/+page.markdoc
```diff
@@ -21,9 +21,13 @@ We use Docker to replicate the production environment for the local deployment o
 
 Use the `appwrite run functions` command to develop your function locally.
 
+| parameter       | Description                                                                                                |
```
Meldiron, line 24: Also ` --user-id <user-id>` to impersonate user, which will automatically set `x-appwrite-user-id`and `x-appwrite-user-jwt` headers on context object.

Meldiron, line 24: Also `  --with-variables`, which sets your production env variables on function that you run locally. Very scary for security, only use it if your production function doesn't have production secrets in env variables, or you are very aware of what this can lead to.

Variables in `.env` of your functio are always loaded. this parameter does not affect it

Meldiron, line 24: Also `--no-reload`. When set, functions will not hot reload. Simply build, start, and keep running. Any changes to your code wont cause restart or anything.

src/routes/docs/products/functions/develop-locally/+page.markdoc
Meldiron, line 80: `You can hot reload your` sounds like you need to do something. Wording along likes `by default, Appwrite CLI hot-reloads your function` might be better

src/routes/docs/products/functions/develop-locally/+page.markdoc
```diff
@@ -71,6 +75,81 @@ You can use headers like dynamic API keys in your function, which give you acces
...
 {% /arrow_link %}
 
+# Hot reloading {% #hot-reloading %}
+
+You can hot reload your locally developed functions, which means you can update the function code and see the changes happen in real time without restarting. Compiled languages must translate the source code into machine code, causing the system to rebuild continuously. However, interpreted languages only rebuild when you change the dependency file. Otherwise, updates can be made through a file swap.
+
+{% table %}
+* &nbsp; {% width=80 %}
+* Language
+* Dependency File
```
Meldiron, line 85: Please double check that list is correct with list in the source code: https://github.com/appwrite/sdk-for-cli/blob/1.6.x/lib/emulation/utils.js#L25

On first sight I can see Node is missing package lock file.

src/routes/docs/products/functions/develop-locally/+page.markdoc
```diff
@@ -21,9 +21,13 @@ We use Docker to replicate the production environment for the local deployment o
 
 Use the `appwrite run functions` command to develop your function locally.
 
+| parameter       | Description                                                                                                |
+|-----------------|------------------------------------------------------------------------------------------------------------|
+| `--port`        | You can set your function port; it defaults to `3000`, or the closest available, i.e. `3001`, `3002`, etc. |
+| `--function-id` | You can autoselect a function, so you don't have to click through the list each time.                      |
```
stnguyen90, line 27: Let's get rid of the "You can" to make it concise and more assertive.

> You can autoselect a function

You're not really auto-selecting because this parameter allows you to manually select the function ahead of time rather than selecting it interactively

src/routes/docs/products/functions/develop-locally/+page.markdoc
```diff
@@ -71,6 +75,81 @@ You can use headers like dynamic API keys in your function, which give you acces
...
+
+{% table %}
+* &nbsp; {% width=80 %}
+* Language
+* Dependency File
+---
+* {% only_dark %}{% icon_image src="/images/platforms/dark/nodejs.svg" alt="Node.js logo" size="m" /%}{% /only_dark %}
+{% only_light %}{% icon_image src="/images/platforms/nodejs.svg" alt="Node.js logo" size="m" /%}{% /only_light %}
+* Node.js
+* `Package.json`
```
stnguyen90, line 90: make sure the case is right or all the files

```suggestion
* `package.json`
```

src/routes/docs/products/functions/develop-locally/+page.markdoc
```diff
@@ -71,6 +75,81 @@ You can use headers like dynamic API keys in your function, which give you acces
...
+---
+* {% only_dark %}{% icon_image src="/images/platforms/dark/ruby.svg" alt="Ruby logo" size="m" /%}{% /only_dark %}
+{% only_light %}{% icon_image src="/images/platforms/ruby.svg" alt="Ruby logo" size="m" /%}{% /only_light %}
+* Ruby
+* `Gemfile`
+---
+* {% only_dark %}{% icon_image src="/images/platforms/dark/deno.svg" alt="Deno logo" size="m" /%}{% /only_dark %}
+{% only_light %}{% icon_image src="/images/platforms/deno.svg" alt="Deno logo" size="m" /%}{% /only_light %}
+* Deno
+* `Import URLs`
```
stnguyen90, line 110: This is not code, so it shouldn't be formatted as code

src/routes/docs/products/functions/develop-locally/+page.markdoc
```diff
@@ -71,6 +75,81 @@ You can use headers like dynamic API keys in your function, which give you acces
 Learn more about dynamic API keys
 {% /arrow_link %}
 
+# Hot reloading {% #hot-reloading %}
```
stnguyen90, line 78: This can just be "Hot reload" instead of "Hot reloading".

In addition, if we're going to add a section, we can remove the sentences mentioning this feature at the bottom of the Develop section.

stnguyen90 (changes requested):

src/routes/docs/products/functions/develop-locally/+page.markdoc
choir27: I'm not sure I see the sentences mentioning hot reloading in the Develop docs or the Local Develop docs. What are you referring to remove here?

src/routes/docs/products/functions/develop-locally/+page.markdoc
choir27: The file swap happens inside the `.appwrite` folder.
- Appwrite puts files from last build (extract it from `tar.gz`)
- Appwrite put files from your function folder there (allows override)
- Appwrite make a new build `tar.gz`
- Appwrite use that to start container again

It's "faking" build step without actually building - just updating files in build file
Regarding Docker container and it's files, Matej believes each container has a folder in /var directory. At least on Linux, no idea on Windows. but thats too low-level. Matej doesn't think we need to explain to this extend though.

src/routes/docs/products/functions/develop-locally/+page.markdoc
```diff
@@ -71,11 +78,86 @@ You can use headers like dynamic API keys in your function, which give you acces
...
+---
+* {% only_dark %}{% icon_image src="/images/platforms/dark/dotnet.svg" alt=".NET logo" size="m" /%}{% /only_dark %}
+{% only_light %}{% icon_image src="/images/platforms/dotnet.svg" alt=".NET logo" size="m" /%}{% /only_light %}
+* .NET
+* .nupkg (NuGet Packages)
+---
+* {% only_dark %}{% icon_image src="/images/platforms/dark/bun.svg" alt="Bun logo" size="m" /%}{% /only_dark %}
+{% only_light %}{% icon_image src="/images/platforms/bun.svg" alt="Bun logo" size="m" /%}{% /only_light %}
+* Bun
+* package.json, package-lock.json, bunlockb
```
stnguyen90, line 138:
```suggestion
* package.json, package-lock.json, bun.lockb
```

src/routes/docs/products/functions/develop-locally/+page.markdoc
```diff
@@ -71,11 +78,86 @@ You can use headers like dynamic API keys in your function, which give you acces
 Learn more about dynamic API keys
 {% /arrow_link %}
 
+# Hot reload {% #hot-reload %}
+
+By default, Appwrite CLI hot-reloads your functions, which means you can update the function code and see the changes happen in real time without restarting. Compiled languages must translate the source code into machine code, causing the system to rebuild continuously. However, interpreted languages only rebuild when you change the dependency file.
```
stnguyen90, line 83: > causing the system to rebuild continuously. However, interpreted languages only rebuild when you change the dependency file. 

More specifically, compiled languages will rebuild on change, but interpreted languages may just restart.

stnguyen90 (changes requested):

src/routes/docs/products/functions/develop-locally/+page.markdoc
```diff
@@ -71,11 +78,86 @@ You can use headers like dynamic API keys in your function, which give you acces
 Learn more about dynamic API keys
 {% /arrow_link %}
 
+# Hot reload {% #hot-reload %}
+
+By default, Appwrite CLI hot-reloads your functions, which means you can update the function code and see the changes happen in real time without restarting. Compiled languages must translate the source code into machine code, causing the system to rebuild continuously. However, interpreted languages only refreshes the function when you change the dependency file.
```
stnguyen90, line 83: > interpreted languages only refreshes the function when you change the dependency file. 

This makes it sound like the function will only update if the dependency file changes, but that's not the case. I also misunderstood what actually happens and where the dependency file comes into play. There are actually 3 cases/scenarios:

1. compiled language + any change
2. interpreted language + change to dependency file
3. interpreted language + change to code

So, this section should look something like:

```suggestion
By default, the Appwrite CLI hot-reloads your functions, which means you can update the function code and the changes will be applied automatically. How this happens differs between runtimes with compiled languages versus interpreted ones.

Because runtimes with compiled languages must translate the source code into machine code, the function must rebuild on change.

When the source code in a runtime with an interpreted languages is updated, the function only needs to restart with the updated file. However, if a dependency file for the interpreted language is updated, the function must rebuild to update the dependencies. Refer to the table below for the dependency files of each language.
```

stnguyen90 (changes requested):

stnguyen90 (approved):

PR #1327, opened 2024-09-10
fix: github link

ArmanNik:
The github link at the bottom of the page now redirects to the correct file

### DIFF
--- a/src/lib/components/Feedback.svelte
+++ b/src/lib/components/Feedback.svelte
@@ -53,7 +53,7 @@
             style="flex-wrap: wrap-reverse;"
         >
             <div class="flex items-center gap-4">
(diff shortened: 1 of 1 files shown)


src/lib/components/Feedback.svelte
```diff
@@ -85,7 +85,7 @@
                     {/if}
                     <li>
                         <a
-                            href="https://github.com/appwrite/website"
+                            href={`https://github.com/appwrite/website/tree/main/src/routes/${$page.url.pathname}`}
```
TorstenDittmann, line 88: this not working with slugs, you should use $page.route.id

TorstenDittmann:
links has double slash `https://github.com/appwrite/website/tree/main/src/routes//docs/products/auth/users`

TorstenDittmann (approved):

PR #1348, opened 2024-09-16
Fix: use native apis to read file contents

ItzNotABug:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)


src/routes/docs/products/functions/develop/+page.markdoc
```diff
@@ -823,7 +823,9 @@ export default async ({ req, res, log }) => {
         case 'json':
             return res.json({"type": "This is a JSON response"});
         case 'binary':
-            return res.binary(InputFile.fromPath('/path/to/file', 'filename'));
+            const bytes = fs.readFileSync('/path/to/file');
+            // or const bytes = await fs.readFile('/path/to/file');
```
Meldiron, line 827: Just one is good, lets go with await approach - better utilizing JS event loop

src/routes/docs/products/functions/develop/+page.markdoc
```diff
@@ -847,7 +848,8 @@ return function ($context) {
         case 'json':
             return $context->res->json(["type" => "This is a JSON response"]);
         case 'binary':
-            return $context->res->binary(InputFile::withPath('file.png'));
+            $fileContent = file_get_contents('file.png');
```
Meldiron, line 851: Let's check with Steven for confirmation. I think it would be better to have a snippet that can do relative path. Because with absolute you would need to `/usr/local/server/src/function/[MY_PATH]`, which is very hard to remember

src/routes/docs/products/functions/develop/+page.markdoc
```diff
@@ -909,7 +913,9 @@ export default async ({ req, res, log }) => {
         case 'json':
             return res.json({type: "This is a JSON response"});
         case 'binary':
-            return res.binary(InputFile.fromPath('/path/to/file.png', 'file.png'));
+            const fileContents = Deno.readFileSync('/path/to/file.png');
```
Meldiron, line 915: Lets be consistent with all file names across languages

src/routes/docs/products/functions/develop/+page.markdoc
```diff
@@ -909,7 +913,9 @@ export default async ({ req, res, log }) => {
...
+            // or const fileContents = await Deno.readFile('/path/to/file.png');
```
Meldiron, line 916: everywhere you gave 2 options, lets always go with the better one. that's always async version

src/routes/docs/products/functions/develop/+page.markdoc
ItzNotABug: > Let's check with Steven for confirmation. I think it would be better to have a snippet that can do relative path. Because with absolute you would need to `/usr/local/server/src/function/[MY_PATH]`, which is very hard to remember

Yep. I think that'd fall on other runtimes too, right?

src/routes/docs/products/functions/develop/+page.markdoc
```diff
@@ -939,7 +945,7 @@ func Main(Context openruntimes.Context) openruntimes.Response {
 	case "json":
 		return Context.Res.Json(map[string]string{"type": "This is a JSON response"})
 	case "binary":
-		file, _ := os.Open("./destiny.png")
+		file, _ := os.Open("./file.png")
```
Meldiron, line 947: In compiled languages not all files are copied, instead, only compilation output.
For this to work, you would need additional command during build step to copy your asset files.
Possible, but to keep it simple we could do: https://gobyexample.com/embed-directive

Let's check other compiled languages and see if they have something similar: C++, Dart, Dotnet, Java/Kotlin, Swift

src/routes/docs/products/functions/develop/+page.markdoc
```diff
@@ -823,7 +823,8 @@ export default async ({ req, res, log }) => {
         case 'json':
             return res.json({"type": "This is a JSON response"});
         case 'binary':
-            return res.binary(InputFile.fromPath('/path/to/file', 'filename'));
+            const bytes = await fs.readFile('/path/to/file');
```
stnguyen90, line 826: Let's make sure we're consistent. If it's `file.png` for the others, let's use that here too.

src/routes/docs/products/functions/develop/+page.markdoc
```diff
@@ -870,7 +870,9 @@ def main(context):
     elif type == 'json':
         return context.res.json({"type": "This is a JSON response"})
     elif type == 'binary':
-        return context.res.binary(InputFile.from_path('file.png'))
+        with open('file.png', 'rb') as file:
+            file_contents = file.read()
+        return context.res.binary(file_contents)
```
stnguyen90, line 875: is `file_contents` still in scope here?

ItzNotABug: It is indeed and accessible to the `context.res` for `binary`.

src/routes/docs/products/functions/develop/+page.markdoc
stnguyen90: This gets very tricky due to how some runtimes will include the assets in the source vs others won't.

The code sample we currently have is essentially relative. We should just stick to that and ensure we use a consistent path for all examples.

Meldiron (approved):

Herry0092 (approved):

PR #1415, opened 2024-10-09
doc-update-apple-indentation-mardoc

raman04-byte:
There should be an update on this apple markdoc as the indentation of the lines is quite important because if not done it throws an error.

## What does this PR do?

It simply update the mardoc as by practical implementation I was facing the issue of the indentation while doing it and it should be clearly visible where to add those lines if someone new who want to try the Appwrite using Apple becomes easier. 

YES


src/routes/docs/quick-starts/apple/+page.markdoc
```diff
@@ -59,19 +59,23 @@ When dependency resolution is complete, click **Add Package** again to add the S
 In order to allow creating OAuth sessions, the following URL scheme must be added to your **Info.plist** file.
 
 ```xml
-<key>CFBundleURLTypes</key>
-<array>
+<plist version="1.0">
```
abnegate, line 62: We don't need the plist definition here, since the intention is to add this block to an existing plist, not create a new one

loks0n:
@raman04-byte I'm going to close this PR, seems like the original is prefered.

Thanks for your efforts!

PR #1513, opened 2024-11-23
Update migrations upgrading docs to avoid confusion

DH-555:
The prior statement was being understood that you need to go from 1.5.0 to 1.6.0, but the reality is that you need to go through the intermediate upgrades/migrations like in version 1.5.10 that fixed some things.

Requiring people to upgrade to the latest patch version solves the issue as they will run the latest, updated migrations of that version prior to going to the next version. 

Going from 1.5.0 to 1.6.0 without doing the 1.5.10 migrations supposes having an unstable instance, usually with broken parts such as functions.


DH-555:
Is it fine/easy to understand?

loks0n (approved):

PR #1561, opened 2024-12-18
feat: document react native oauth

loks0n:


src/routes/docs/products/auth/oauth2/+page.markdoc
```diff
@@ -118,6 +118,37 @@ try await account.createOAuth2Session(
...
+const account = new Account(client);
+
+// Go to OAuth provider login page
+const loginUrl = await account.createOAuth2Token(
+    OAuthProvider.Github, // provider
+    Linking.createURL('/success'),
+    Linking.createURL('/failed'),
+    ['repo', 'user'] // scopes (optional)
+);
+const result = WebBrowser.openAuthSessionAsync(loginUrl);
```
stnguyen90, line 141:
```suggestion
const result = await WebBrowser.openAuthSessionAsync(loginUrl);
```

src/routes/docs/products/auth/oauth2/+page.markdoc
```diff
@@ -118,6 +118,37 @@ try await account.createOAuth2Session(
...
+
+// On your `/success` route, create a session from the result
+const url = new URL(result.url);
+const secret = url.searchParams.get("secret");
+const userId = url.searchParams.get("userId");
+
+await account.createSession(userId, secret);
```
stnguyen90, line 148: There's no need to create an actual route. the code execution continues after `openAuthSessionAsync()`

src/routes/docs/products/auth/oauth2/+page.markdoc
```diff
@@ -118,6 +118,37 @@ try await account.createOAuth2Session(
...
+    .setEndpoint('https://cloud.appwrite.io/v1') // Your API Endpoint
+    .setProject('<PROJECT_ID>');                 // Your project ID
+
+const account = new Account(client);
+
+// Go to OAuth provider login page
+const loginUrl = await account.createOAuth2Token(
+    OAuthProvider.Github, // provider
+    Linking.createURL('/success'),
+    Linking.createURL('/failed'),
```
stnguyen90, line 138: The really difficult part is what gets generated from this.

According to the [docs](https://docs.expo.dev/versions/latest/sdk/linking/#linkingcreateurlpath-namedparameters):
* Development and production builds: <scheme>://path - uses the optional scheme property if provided, and otherwise uses the first scheme defined by your app config
* Web (dev): https://localhost:19006/path
* Web (prod): https://myapp.com/path
* Expo Go (dev): exp://128.0.0.1:8081/--/path

So, when you're developing locally with expo go, you need to add a web platform with 128.0.0.1 (or whatever your LAN IP is. However, development and production builds, don't have a hostname at all so that's going to fail the validation for the success and failure params.

stnguyen90 (changes requested):

src/routes/docs/products/auth/oauth2/+page.markdoc
stnguyen90: To ensure things work for both expo go and development/production builds, I had to do this:

```javascript
const LoginScreen = () => {
    const router = useRouter();
    
    const deepLink = new URL(Linking.createURL("login-github", { isTripleSlashed: true }));
    console.log(deepLink);
    if (!deepLink.hostname) {
        deepLink.hostname = "localhost";
    }
    console.log(deepLink);

    const handleLogin = async () => {
        try {
            const success = `${deepLink}/success`;
            const failure = `${deepLink}/failure`;
            console.log(success);
            const response = await account.createOAuth2Token(
                OAuthProvider.Google,
                success,
                failure
            );
            console.log(response);
            let newURL = response.href.toString();
            const result = await WebBrowser.openAuthSessionAsync(newURL);
            console.log(result);

            if (result.type !== "success") {
                // handle error for if they cancelled the login
            }

            // extract secret query string param from url
            const url = new URL(result.url);

            if (url.pathname !== "/login-github/success") {
                // handle error if session wasn't created successfully by Appwrite
            }

            const secret = url.searchParams.get("secret");
            const userId = url.searchParams.get("userId");
            await account.createSession(userId, secret);
            const user = await account.get();
            console.log(user);
            // redirect to profile after successful sign in
            router.push("/profile");
        } catch (error) {
            console.log(error);
        }
    };

    return (
        <View style={styles.container}>
            <TouchableOpacity onPress={handleLogin} style={styles.button}>
                <Text style={styles.buttonText}>Login with Google</Text>
            </TouchableOpacity>
        </View>
    );
};
```

stnguyen90 (approved):

PR #1771, opened 2025-02-13
fix tabs cutting off last item

thejessewinton:
## What does this PR do?

Fixes tabs on docs cutting off last item.

## Test Plan

## Related PRs and Issues

(Write your answer here.)


src/markdoc/tags/Tabs.svelte
ItzNotABug, line 1: @thejessewinton can we keep the tab item text on the same line? Here's the current state after this change -

Screenshot 2025-02-17 at 11 26 11 AM

ItzNotABug (changes requested):

src/markdoc/tags/Tabs.svelte
thejessewinton: @ItzNotABug Yeah, I'll add a line-clamp to the select label so, ideally, the items that aren't in the dropdown are fully rendered.

thejessewinton: Fixed @ItzNotABug. This is the only real way this layout works correctly without throwing weird layout issues- the select overflows, but the container is scrollable.

Screenshot 2025-02-20 at 10 06 56 AM

ItzNotABug: Even though its scrollable, on first glance it does look broken, just by the ui of it. What if, we remove the text `router` for this usecase? Description anyway mentions its related to router like `Vue router` & `React router`. We could just keep the title as React, Vue, etc. should be ok.

ItzNotABug (approved):

DH-555 (approved):

PR #1804, opened 2025-02-26
Add blog post on user role guests error

ebenezerdon:
## What does this PR do?
Adds new blog post

## Test Plan
- `/blog/post/user-role-guests-missing-scope-account`


src/routes/blog/post/user-role-guests-missing-scope-account/+page.markdoc
```diff
@@ -0,0 +1,118 @@
...
+
+For local development:
+
+- Allow 3rd party cookies in your browser settings
+- Some browsers have specific development settings to allow cross-domain cookies during local development
+
+For production:
+
+- [Configure a custom domain](https://appwrite.io/docs/advanced/platform/custom-domains) for your Appwrite project so that cookies are sent from the same domain as your application
+- This ensures that session cookies are properly sent with each request
```
stnguyen90, line 72: These bullet points makes it seem like there are 2 different options but that's not really the case 🧐

src/routes/blog/post/user-role-guests-missing-scope-account/+page.markdoc
```diff
@@ -0,0 +1,118 @@
...
+
+## Option 2: Use OAuth2 tokens
+
+Another approach is to use Appwrite's OAuth2 token-based authentication, which bypasses third-party cookie restrictions entirely. Instead of `createOAuth2Session()`, you would use `createOAuth2Token()` which implements a more privacy-friendly flow that works in all browsers.
```
stnguyen90, line 76: > implements a more privacy-friendly flow that works in all browsers.

I wouldn't really call it a more privacy-friendly flow. It's just this token approach will use local storage as a fallback for cases where cookies won't work.

src/routes/blog/post/user-role-guests-missing-scope-account/+page.markdoc
```diff
@@ -0,0 +1,118 @@
...
+
+For a detailed implementation guide, check out this article on [Fixing OAuth2 authentication issues in Appwrite Cloud](https://appwrite.io/blog/post/fixing-oauth2-issues-in-appwrite-cloud?doFollow=true), which covers this approach in depth.
+
+Keep in mind that non-public files still require session cookies, so stick with Option 1 if your app needs to handle such files.
```
stnguyen90, line 80: > so stick with Option 1 if your app needs to handle such files.

You can still go with option 2, but you'll need to set up custom domains so the cookie is a first party cookie.

src/routes/blog/post/user-role-guests-missing-scope-account/+page.markdoc
ebenezerdon: That makes sense. Updated

src/routes/blog/post/user-role-guests-missing-scope-account/+page.markdoc
ebenezerdon: Updated

src/routes/blog/post/user-role-guests-missing-scope-account/+page.markdoc
ebenezerdon: Updated

stnguyen90 (approved):

PR #1805, opened 2025-02-27
Feat: add sentry release environment variable

ernstmul:
## What does this PR do?
Set release tag name for Sentry

✅


Dockerfile
```diff
@@ -45,6 +45,9 @@ ENV GITHUB_TOKEN ${GITHUB_TOKEN}
 ARG SENTRY_AUTH_TOKEN
 ENV SENTRY_AUTH_TOKEN ${SENTRY_AUTH_TOKEN}
 
+ARG SENTRY_RELEASE
```
ItzNotABug, line 48: will this be needed on coolify as well? will the build go okay without it or will it crash? default value?

ernstmul: Do we have the other sentry keys in coolify as well? Do we want those test envs in Sentry at all?

stnguyen90: It should default to some random hash.

stnguyen90 (approved):

PR #1815, opened 2025-03-04
Feat new events

ernstmul:
## What does this PR do?
Add new analytic events

✅


src/routes/+layout.svelte
```diff
@@ -118,8 +125,31 @@
...
+        const scrollY = window.scrollY;
+        const docHeight = document.documentElement.scrollHeight - window.innerHeight;
+        const scrollPercentage = scrollY / docHeight;
+
+        thresholds.forEach((threshold) => {
+            if (scrollPercentage >= threshold && !tracked.has(threshold)) {
+                const pageName =
+                    location.pathname.slice(1) === ''
+                        ? 'home'
+                        : location.pathname.slice(1).replace(/\//g, '-');
```
ItzNotABug, line 139: would `$page` store be better here?

src/routes/+layout.svelte
```diff
@@ -118,8 +125,31 @@
...
+
+                const eventName = `${pageName}_scroll-depth_${threshold * 100}prct_scroll`;
+                tracked.add(threshold);
+                trackEvent({
+                    plausible: { name: eventName },
+                    posthog: { name: eventName }
+                });
```
ItzNotABug, line 146: what do you think of adding some `debounce`?

ItzNotABug (changes requested):

src/routes/+layout.svelte
ernstmul: Isn't needed, it only fires once per page load. It keeps track of the thresholds being passed, so it doesn't fire again!

src/routes/+layout.svelte
ernstmul: Adjsuted!

ItzNotABug (approved):

PR #1868, opened 2025-03-18
Update queries docs page to have section for each operator

ebenezerdon:
## What does this PR do?
Currently, all operators are crumbled into a table. This changes it so we have a section for each operator. 

## Test Plan
- `/docs/products/databases/queries`

## Related PRs and Issues
DRL-1351


stnguyen90 (approved):
Can we add the REST API format too? Maybe as a follow up PR?

PR #1969, opened 2025-05-12
Add RLS reference to document level security documentation

ebenezerdon:
## What does this PR do?
- Adds an info box to the database permissions documentation explaining how Appwrite's document-level security is similar to Row Level Security (RLS) in other database systems.

## Test Plan
- `/docs/products/databases/permissions`

### DIFF
--- a/src/routes/docs/products/databases/permissions/+page.markdoc
+++ b/src/routes/docs/products/databases/permissions/+page.markdoc
@@ -23,6 +23,10 @@ Configure collection level permissions by navigating to **Your collection** > **
 Document level permissions grant access to individual documents.
 If a user has read, create, update, or delete permissions at the document level, the user can access the **individual document**.
 
+{% info title="Does Appwrite support Row Level Security (RLS)?" %}
+Document level security in Appwrite is similar to what is known as Row Level Security (RLS) found in database systems like PostgreSQL. Both concepts allow you to control access to individual records based on user identity and roles. If you're familiar with RLS, you'll find Appwrite's document level security provides similar granular access control.
+{% /info %}
+
 Document level permissions are only applied if Document Security is enabled in the settings of your collection.
 Enable document level permissions by navigating to **Your collection** > **Settings** > **Document security**.
 


src/routes/docs/products/databases/permissions/+page.markdoc
```diff
@@ -23,6 +23,10 @@ Configure collection level permissions by navigating to **Your collection** > **
 Document level permissions grant access to individual documents.
 If a user has read, create, update, or delete permissions at the document level, the user can access the **individual document**.
 
+{% info title="Does Appwrite support Row Level Security (RLS)?" %}
+Document level security in Appwrite is similar to what is known as Row Level Security (RLS) found in database systems like PostgreSQL. Both concepts allow you to control access to individual records based on user identity and roles. If you're familiar with RLS, you'll find Appwrite's document level security provides similar granular access control.
```
stnguyen90, line 27: > provides similar granular access control.

While both allows giving granular access, RLS is quite different than Appwrite's permission model so we should make sure that's clear.

ebenezerdon: @stnguyen90 Can you help with the specific difference(s) so we can make it clearer?

stnguyen90: Postgres' RLS allows you to add a policy to a table. For example, given table:

```sql
CREATE TABLE accounts (manager text, company text, contact_email text);
```

this policy:

```sql
CREATE POLICY account_managers ON accounts TO managers
    USING (manager = current_user);
```

Makes it so when a query is done on the `accounts` table, only rows where the `manager` column matches `current_user` will return.

With Appwrite's document-level permissions, you set on the document which user should have access to that document.

Does that give you enough information to better word it?

stnguyen90 (commented):
Left a [comment](https://github.com/appwrite/website/pull/1969#discussion_r2181056397).

src/routes/docs/products/databases/permissions/+page.markdoc
ebenezerdon: Yes, that makes sense @stnguyen90. What do you think about the update I made?

src/routes/docs/products/databases/permissions/+page.markdoc
```diff
@@ -23,6 +23,10 @@ Configure collection level permissions by navigating to **Your collection** > **
 Document level permissions grant access to individual documents.
 If a user has read, create, update, or delete permissions at the document level, the user can access the **individual document**.
 
+{% info title="Does Appwrite support Row Level Security (RLS)?" %}
+Document level security in Appwrite is similar to Row Level Security (RLS) found in database systems like PostgreSQL, as both provide granular access control to individual records. However, they work differently: RLS uses policies that automatically filter results based on conditions, while Appwrite requires explicitly setting permissions on each document.
```
stnguyen90, line 27: > _explicitly_ setting permissions

Does this sound negative to you? Can you reword it to not sound like a drawback/flaw?

stnguyen90 (changes requested):

PR #2078, opened 2025-06-07
feat: generate llms.txt on build

TorstenDittmann:


EVDOG4LIFE:
@TorstenDittmann 

I'm open to building at build time. My primary concern would be the size if we are taking the content from each page - if we just build the header content (url + article title) that would be maintainable as the full content might nuke the context window. 

We could create a llms.txt with just the structure and then a llms-full.txt with the full dump.

PR #2137, opened 2025-06-23
fix(docs): header display breakpoints and client side routing

TorstenDittmann:
fixed tablet viewport by adding the navigation back
https://github.com/user-attachments/assets/77421790-7db6-4130-b694-c1a31e1a1bb5

fixed client side routing
https://github.com/user-attachments/assets/60520aaa-1885-4c8b-82d1-c51925294496


thejessewinton (commented):
LGTM

PR #2151, opened 2025-06-27
Clarify userId param in endpoints that create accounts

ebenezerdon:
## What does this PR do?
Update auth doc to clarify account creation process for email and phone SMS methods

## Related PRs and Issues
DRL-1419

### DIFF
--- a/src/routes/docs/products/auth/email-otp/+page.markdoc
+++ b/src/routes/docs/products/auth/email-otp/+page.markdoc
@@ -22,7 +22,7 @@ Both allow passwordless login flows with different advantages.
 
 Email OTP authentication is done using a two-step authentication process. The authentication request is initiated from the client application and an email message is sent to the user's email inbox. The email will contain a 6-digit number the user can use to log in.
 
-Send an an email to initiate the authentication process. A **new account** will be created for this email if it has never been used before.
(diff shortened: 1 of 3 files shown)


src/routes/docs/products/auth/email-otp/+page.markdoc
```diff
@@ -22,7 +22,7 @@ Both allow passwordless login flows with different advantages.
 
 Email OTP authentication is done using a two-step authentication process. The authentication request is initiated from the client application and an email message is sent to the user's email inbox. The email will contain a 6-digit number the user can use to log in.
 
-Send an an email to initiate the authentication process. A **new account** will be created for this email if it has never been used before.
+Send an an email to initiate the authentication process. If the email has never been used, a **new account is generated** using the provided `userId`, then the user will receive an email. If the email is already attached to an account, the **user ID is ignored** and the user will receive an email with the one-time password.
```
stnguyen90: Sounds odd to say a new account is generated. Also part of it sounds like a run on.

```suggestion
Send an email to initiate the authentication process. If the email address has never been used, a **new account is created** using the provided `userId`. Otherwise, if the email address is already attached to an account, the **user ID is ignored**. Then, the user will receive an email with the one-time password.
```

stnguyen90 (changes requested):
Left a comment that can be applied to all 3 pages.

ebenezerdon:
> Left a comment that can be applied to all 3 pages.

Thanks, updated

Herry0092:
#2151 is now merged into main. Next step: I’ll test the deployment and
update the platform docs.

PR #2186, opened 2025-07-15
feat:added a line in free pricing component

HarshMN2345:
### Add pricing page note

This PR adds the following text to the **Free** plan section at the bottom of the pricing page:

After: 
image

### DIFF
--- a/src/routes/pricing/+page.svelte
+++ b/src/routes/pricing/+page.svelte
@@ -141,7 +141,9 @@
                                                     >
                                                 </li>
                                             </ul>
-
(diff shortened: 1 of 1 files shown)


src/routes/pricing/+page.svelte
ItzNotABug, line 1: 1. As per design, the divider seems to be missing.
2. The price comparison table should also include limited projects and pausing details. Please check with the design team.
3. Confirm if we need to show some sort of tooltip around this here and under the comparison table for some more explanation.

PR #2192, opened 2025-07-16
Fix education page pfp issue

HarshMN2345:
## What does this PR do?

Solves the truncation of the pfp issue that I was facing in some of the mobile devices 
## Before
image
## After 
image

## Test Plan

## Related PRs and Issues

yes


src/lib/components/CommunitySupportChat.svelte
```diff
@@ -87,9 +91,51 @@
 </ul>
 
 <style>
+    .web-user-box-image {
```
ItzNotABug, line 90: why do we need this many props with this custom css when the `style="..."` already has the similar specs? Can you confirm which do we want here.

ItzNotABug (changes requested):

src/lib/components/CommunitySupportChat.svelte
```diff
@@ -87,6 +87,12 @@
...
+        width: 40px !important;
+        height: 40px !important;
+        min-width: 40px !important;
+        min-height: 40px !important;
```
ItzNotABug: can you verify if we need all of these `!important` tags?

src/lib/components/CommunitySupportChat.svelte
```diff
@@ -88,10 +88,10 @@
...
 <style>
     .web-user-box-image {
-        width: 40px !important;
-        height: 40px !important;
-        min-width: 40px !important;
-        min-height: 40px !important;
+        width: 40px;
+        height: 40px;
+        min-width: 40px;
+        min-height: 40px;
```
ItzNotABug, line 94: all values are same, we might not need the `min-` props in that case here.

Screenshot 2025-07-17 at 1 05 05 PM

PR #2409, opened 2025-09-16
fix: rendering of multi method sdk specs

ChiragAgg5k:
## What does this PR do?

skips deprecated methods loading in case of multi methods + fixes rendering of method name (if `desc` is empty show summary)

## Test Plan
Screenshot 2025-09-16 at 4 44 15 PM

yes.

## Summary by CodeRabbit

* **New Features**
  * Added deprecation metadata so deprecated variants are hidden from outputs.
  * Improved handling of alternate/added method variants for GET/POST/PUT/PATCH, with clearer summaries and merged demo/metadata.
  * Responses for added variants now include the primary response model.

* **Bug Fixes**
  * Request bodies for added variants now reflect declared parameters accurately.
  * DELETE variants no longer produce confusing alternate entries when other variants exist.

### DIFF
--- a/src/lib/utils/specs.ts
+++ b/src/lib/utils/specs.ts
@@ -32,6 +32,11 @@ type SDKMethodModel = {
     name: string;
 };
 
+type AppwriteDeprecated = {
+    since: string;
+    replaceWith: string;
+};
+
 type AppwriteOperationObject = OpenAPIV3.OperationObject & {
     'x-appwrite': {
         method: string;
@@ -60,6 +65,7 @@ type AppwriteAdditionalMethod = {
     responses: { code: number; model: string }[];
     description: string;
     demo: string;
+    deprecated?: AppwriteDeprecated;
 };
 
 export type AppwriteSchemaObject = OpenAPIV3.SchemaObject & {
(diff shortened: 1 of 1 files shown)


src/lib/utils/specs.ts
```diff
@@ -216,11 +222,16 @@ function* iterateAllMethods(
...
                 yield {
                     method: OpenAPIV3.HttpMethods.POST,
                     value: {
                         ...methods.post,
-                        summary: additionalMethod.desc,
+                        summary:
+                            additionalMethod.desc.length > 0
+                                ? additionalMethod.desc
+                                : methods.post.summary,
                         description: additionalMethod.description,
```
abnegate: Does this assume additionalMethods will only ever be `POST`?

ChiragAgg5k: actually, we should have this replicated for all other too

src/lib/utils/specs.ts
```diff
@@ -169,6 +175,71 @@ function filterRequestBodyProperties(
...
+): Generator<{
+    method: OpenAPIV3.HttpMethods;
+    value: OpenAPIV3.OperationObject | AppwriteOperationObject;
+    url: string;
+}> {
+    const appwriteMethod = method as AppwriteOperationObject;
+    const additionalMethods = appwriteMethod['x-appwrite'].methods!;
+
+    for (const additionalMethod of additionalMethods) {
+        if (additionalMethod.deprecated) continue;
```
abnegate, line 208: I think we also need to ensure the count of additional methods is > 1 here

src/lib/utils/specs.ts
```diff
@@ -169,6 +175,71 @@ function filterRequestBodyProperties(
...
+                responses: {
+                    ...appwriteMethod.responses,
+                    [additionalMethod.responses[0].code]: {
+                        content: {
+                            'application/json': {
+                                schema: {
+                                    $ref: additionalMethod.responses[0].model
+                                }
+                            }
+                        }
```
abnegate: This won't work for deletes, for those there should only be `"description": "No content"`

src/lib/utils/specs.ts
ChiragAgg5k: we are already in additionalMethod loop, here we see the object of each additionalMethod. if deperecated is not set, its simply undefined

abnegate: Will it still work if there are multiple methods, but all are deprecated?

abnegate (approved):

PR #2525, opened 2025-10-26
fix: dynamic mobile nav positioning for banner height

HarshMN2345:
## What does this PR do?

## Test Plan

## Related PRs and Issues

(Write your answer here.)

## Summary by CodeRabbit

* **Bug Fixes**
  * Mobile navigation now correctly offsets from the top when a top banner is present, preventing overlap with banner content.
  * Navigation positioning recalculates on window resize and still closes appropriately during navigation or when resized.

### DIFF
--- a/src/lib/components/MobileNav.svelte
+++ b/src/lib/components/MobileNav.svelte
@@ -36,9 +36,7 @@
         }
     }
 
-    onMount(() => {
-        updateBannerHeight();
-    });
+    onMount(updateBannerHeight);
 </script>
 
 <svelte:window


src/lib/components/MobileNav.svelte
```diff
@@ -7,22 +7,40 @@
...
+    function updateBannerHeight() {
+        if (browser) {
+            const banner = document.getElementById('hackathon-banner');
+            bannerHeight = banner ? banner.getBoundingClientRect().height : 0;
+        }
+    }
+
+    onMount(() => {
+        updateBannerHeight();
+        window.addEventListener('resize', updateBannerHeight);
```
ItzNotABug: we already have `svelte:window`, lets use that to avoid managing the adding/removing listener 👍. Svelte automatically takes care of removing listeners.

ItzNotABug (changes requested):

src/lib/components/MobileNav.svelte
```diff
@@ -7,22 +7,43 @@
...
+    let bannerHeight = 0;
+
     afterNavigate(() => {
         open = false;
         document.body.style.overflow = '';
     });
+
+    function updateBannerHeight() {
+        if (browser) {
+            const banner = document.getElementById('hackathon-banner');
```
ItzNotABug: we can also cache this assuming the banner is cancellable instead of always fetching the element. something like -
```ts
let banner: HtmlElement | null = null;

function getBannerElement() {
  if (banner !== null) return banner;

  banner = document.getElementById('hackathon-banner');
  if (banner) {
    return banner;
  } else {
    return null;
  }
}
```
then just reference `banner` in the `updateBannerHeight` with a null check /guard because the banner won't necessarily be there always!

Not needed if the banner can be removed via some interaction!

src/lib/layouts/Main.svelte
```diff
@@ -162,7 +162,7 @@
 
 <div class="relative contents h-full">
     {#if !page.url.pathname.includes('/init')}
-        <div class="border-smooth relative z-10 border-b bg-[#19191C]">
+        <div class="border-smooth relative z-10 border-b bg-[#19191C]" id="hackathon-banner">
```
ItzNotABug: not necessary that this is always a `hackathon` banner, we could use an agnostic name for it.

src/lib/components/MobileNav.svelte
```diff
@@ -7,22 +7,43 @@
...
+        open && (open = false);
+        updateBannerHeight();
+    }}
+/>
 
 <nav
-    class="web-side-nav sticky top-[73px] block max-h-screen overflow-hidden lg:hidden"
+    class="web-side-nav sticky block max-h-screen overflow-hidden lg:hidden"
     class:hidden={!open}
+    style="top: calc(73px + {bannerHeight}px);"
```
ItzNotABug:
```ts
style:top="..."
```
svelte style directive!

src/lib/components/MobileNav.svelte
```diff
@@ -7,22 +7,51 @@
...
+    function updateBannerHeight() {
+        if (browser) {
+            const bannerElement = getBannerElement();
+            bannerHeight = bannerElement ? bannerElement.getBoundingClientRect().height : 0;
+        }
+    }
+
+    onMount(() => {
+        updateBannerHeight();
+    });
```
ItzNotABug:
```suggestion
    onMount(updateBannerHeight);
```

ItzNotABug (approved):

PR #2615, opened 2025-11-18
add next outputs blog

atharvadeosthale:
## Summary by CodeRabbit

* **Documentation**
  * Published a new blog article comparing Next.js output modes (default, export, standalone). Covers how each mode behaves, benchmark and build-size observations, performance trade-offs, practical recommendations for choosing standalone vs normal mode, and guidance for selecting the right approach with links for further reading.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/src/routes/blog/post/nextjs-output-modes/+page.markdoc
+++ b/src/routes/blog/post/nextjs-output-modes/+page.markdoc
@@ -2,7 +2,7 @@
 layout: post
 title: Next.js output modes — Standalone vs Default build
 description: Learn about different output modes in Next.js and find out which of them would be the best for you.
-date: 2025-12-11
+date: 2025-12-12
 cover: /images/blog/nextjs-output-modes/cover.png
 timeToRead: 5
 author: atharva
@@ -26,7 +26,7 @@ It's important to note that if you remove `node_modules`, which usually is the b
 
 The default mode is typically deployed by running `next start`. This is convenient, but gives you less control over the underlying server process than approaches where you own the server entrypoint (e.g., injecting custom server logic/handlers as part of your runtime).
 
-You might already think, shipping `node_modules` is a very big size hit, and it's true, but not always a bad thing. We will look at why later.
+You might already think that shipping `node_modules` takes a big hit on storage, and while it's true, but not always a bad thing. We will later look at why.
 
 # The "export" mode
 
@@ -48,14 +48,14 @@ Even with all the benefits, there's a very good reason for this mode not being t
 
 # Our observations
 
-I created a [benchmark](https://github.com/appwrite-community/nextjs-outputs-bench) to check the sizes of the builds. I had four Next.js apps and ran `next build` on them:
+We prepared a [benchmark](https://github.com/appwrite-community/nextjs-outputs-bench) where we ran the `next build` command for four Next.js apps to check how large their build sizes were:
 
 1. Default mode, no dependencies
 2. Standalone mode, no dependencies
 3. Default mode, very heavy dependencies
 4. Standalone mode, very heavy dependencies
 
-Then I compared the sizes of these builds:
+Then, we compared the sizes of these builds and observed the following:
 
 ```
 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
@@ -77,7 +77,7 @@ Then I compared the sizes of these builds:
      Standalone mode: +0 KB (0.0% increase)
 ```
 
-The result is clear: standalone mode wins as it's consistently shown lower build sizes. Interestingly, same build size because even though we install heavy dependencies, we aren't actually importing/using them at all in the apps, so these dependencies aren't even getting traced and bundled, which is ideal! Lower build sizes mean faster cold starts on your app.
+The result was clear: standalone mode consistently showed lower build sizes. Interestingly, even when we installed heavy dependencies in the app, we observed that the build size for the apps with standalone mode remained unchanged. Because those dependencies were not being imported and used in the app, they were not bundled, which is ideal. Lower build sizes also result in faster cold starts for your app, which creates a better user experience.
 
 # Which is the best for me?
 


tessamero:
Two things - it's not really a tutorial so does it make sense to update the category?

Also, can we use a different word than "funnily" as it throws off the content

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,85 @@
+---
+layout: post
+title: Next.js output modes — Standalone vs Normal build
```
Meldiron: Lets double check what they call the "normal build", worried it might not be accurate term. if we cant find any, "default" might be better.

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,85 @@
...
+timeToRead: 5
+author: atharva
+category: tutorial
+featured: false
+---
+
+Next.js is a very popular React framework. People choose it because it blends in with the server very well, using both client-side and server-side capabilities to provide a better user experience. Next.js is packed with cutting-edge React features like RSC. However, such complexity also brings complexity in deploying your app.
+
+Appwrite Sites enables you to deploy your Next.js app seamlessly without any hassle of configuration. Appwrite handles all the complexity involved in deploying such complex apps, so that you can focus on the core logic of your app.
+
```
Meldiron, line 16: I just read complex 4 times. I think thats too much

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,85 @@
...
+You might already think, shipping `node_modules` is a very big size hit, and it's true, but not always a bad thing. We will look at why later.
+
+# Standalone mode
+
+When building your app in standalone mode, Next.js traces all the files that are required by your app from your `node_modules`. This means that only the functionality your app uses will be bundled into your Next.js build. In this case, you can exclude `node_modules` when calculating sizes, as the build doesn't rely on `node_modules` as it already bundles all necessary code.
+
+This results in bigger `.next` folder, but we also have to factor in the fact that `node_modules` are essential for the default mode, and not the standalone mode.
+
+However, there's a very good reason for this mode not being the default yet.
+
```
Meldiron, line 48: There is also static mode (output: export). And in there  a few configurations useful specifically or SPA too (like railing shash,

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,85 @@
...
+In this article, we will talk about the different Next.js output modes, what would be better for you, and what we have observed when using different output modes with Appwrite Sites.
+
+# The "default" mode
+
+The default mode relies on your dependencies stored in your `node_modules` and the build bundle stored in `.next` folder after running `next build`. To run your app in this output mode, you should run `next start` to serve your app in production.
+
+It's important to note that if you remove `node_modules`, which usually is the biggest chunk of your app, your app won't run anymore, as your builds also rely on `node_modules` to work properly. So, when counting the size of your app, you should also count in `node_modules` folder as well.
+
+You might already think, shipping `node_modules` is a very big size hit, and it's true, but not always a bad thing. We will look at why later.
+
```
Meldiron, line 30: Another big part is "magic box" approach - nuxt start does something which you dont really understand. and you can configure some stuff, like port, but you cannot configure others, like custom middleware or global error handling.

You can do those things directly in next.js, but still, some setups, like appwite/website, benefits from ability to inject some scripts into HTTP listener itself.

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,85 @@
...
+# Standalone mode
+
+When building your app in standalone mode, Next.js traces all the files that are required by your app from your `node_modules`. This means that only the functionality your app uses will be bundled into your Next.js build. In this case, you can exclude `node_modules` when calculating sizes, as the build doesn't rely on `node_modules` as it already bundles all necessary code.
+
+This results in bigger `.next` folder, but we also have to factor in the fact that `node_modules` are essential for the default mode, and not the standalone mode.
```
Meldiron: From what I remember, you dont need to ship entire .next folder, you only need `.nex/standalone`.

And there is downside - this doesnt bundle static assets, as they recommend to deploy them on CDN or S3 instead.
So for minimal setup, you need a few extra commands to copy static assets into this standalone directory.

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,85 @@
...
+     Default:      712.48 MB (.next + node_modules)
+     Standalone:   57.84 MB
+     📉 Difference:  654.64 MB smaller
+
+  🔹 Impact of Heavy Dependencies
+     Default mode:    +281.06 MB (65.1% increase)
+     Standalone mode: +0 KB (0.0% increase)
+```
+
+The result is clear: standalone mode wins as it's consistently shown lower build sizes. Funnily same build size because even though we install heavy dependencies, we aren't actually importing/using them at all in the apps, so these dependencies aren't even getting traced and bundled, which is ideal! Lower build sizes mean faster cold starts on your app.
```
Meldiron: I would avoid term `funily`

static/images/blog/nextjs-output-modes/cover.png
Meldiron, line 1: Wrong image

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,85 @@
...
+1. Normal mode, no dependencies
+2. Standalone mode, no dependencies
+3. Normal mode, very heavy dependencies
+4. Standalone mode, very heavy dependencies
+
+Then I compared the sizes of these builds:
+
+```
+━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
+  📈 Summary & Comparisons
```
Meldiron, line 62: Lets also below memtion benefit of smaller build. Faster cold-starts, less storage cost, faster bundling and distribution... 

And I think faster builds too, but let's benchmark that first

Meldiron: When benchamrking buiilds, let'ß benchmark both cached (with .next from previosu build), and uncached

atharvadeosthale: Not faster builds, tried the bench, the builds are a bit slower on standalone

atharvadeosthale:
> Two things - it's not really a tutorial so does it make sense to update the category?
> 
> Also, can we use a different word than "funnily" as it throws off the content

No other category fits in currently.

And removed "funnily"!

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,98 @@
...
+
+This results in bigger `.next/standalone` folder, but we also have to factor in the fact that `node_modules` are essential for the default mode, and not the standalone mode.
+
+The `.next/standalone` folder does not contain any static assets, and Next.js recommends serving them through a CDN. For a minimal setup, you can run a few commands to copy static assets into the standalone directory.
+
+Even with all the benefits, there's a very good reason for this mode not being the default yet.
+
+# Our observations
+
+I created a [benchmark](https://github.com/atharvadeosthale/nextjs-outputs-bench/tree/main) to check the sizes of the builds. I had four Next.js apps and ran `next build` on them:
```
adityaoberai: Can you shift this to the appwrite-community GitHub org?

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,98 @@
+---
+layout: post
+title: Next.js output modes — Standalone vs Default build
+description: Learn about different output modes in Next.js and find out which of them would be the best for you.
+date: 2025-12-11
```
atharvadeosthale:
```suggestion
date: 2025-12-12
```

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,98 @@
...
+# Our observations
+
+I created a [benchmark](https://github.com/appwrite-community/nextjs-outputs-bench) to check the sizes of the builds. I had four Next.js apps and ran `next build` on them:
+
+1. Default mode, no dependencies
+2. Standalone mode, no dependencies
+3. Default mode, very heavy dependencies
+4. Standalone mode, very heavy dependencies
+
+Then I compared the sizes of these builds:
```
adityaoberai:
```suggestion
We prepared a [benchmark](https://github.com/appwrite-community/nextjs-outputs-bench) where we ran the `next build` command for four Next.js apps to check how large their build sizes were:

1. Default mode, no dependencies
2. Standalone mode, no dependencies
3. Default mode, very heavy dependencies
4. Standalone mode, very heavy dependencies

Then, we compared the sizes of these builds and observed the following:
```

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,98 @@
...
+     Default:      712.48 MB (.next + node_modules)
+     Standalone:   57.84 MB
+     📉 Difference:  654.64 MB smaller
+
+  🔹 Impact of Heavy Dependencies
+     Default mode:    +281.06 MB (65.1% increase)
+     Standalone mode: +0 KB (0.0% increase)
+```
+
+The result is clear: standalone mode wins as it's consistently shown lower build sizes. Interestingly, same build size because even though we install heavy dependencies, we aren't actually importing/using them at all in the apps, so these dependencies aren't even getting traced and bundled, which is ideal! Lower build sizes mean faster cold starts on your app.
```
adityaoberai:
```suggestion
The result was clear: standalone mode consistently showed lower build sizes. Interestingly, even when we installed heavy dependencies in the app, we observed that the build size for the apps with standalone mode remained unchanged. Because those dependencies were not being imported and used in the app, they were not bundled, which is ideal. Lower build sizes also result in faster cold starts for your app, which creates a better user experience.
```

src/routes/blog/post/nextjs-output-modes/+page.markdoc
```diff
@@ -0,0 +1,98 @@
...
+
+# The "default" mode
+
+The default mode relies on your dependencies stored in your `node_modules` and the build bundle stored in `.next` folder after running `next build`. To run your app in this output mode, you should run `next start` to serve your app in production.
+
+It's important to note that if you remove `node_modules`, which usually is the biggest chunk of your app, your app won't run anymore, as your builds also rely on `node_modules` to work properly. So, when counting the size of your app, you should also count in `node_modules` folder as well.
+
+The default mode applies a "magic box" approach. The command `next start` does some *magic* to serve your app. This removes a lot of *configurability* from your app, like custom middlewares or global error handling. The other modes allow some more configurability.
+
+You might already think, shipping `node_modules` is a very big size hit, and it's true, but not always a bad thing. We will look at why later.
```
adityaoberai:
```suggestion
You might already think that shipping `node_modules` takes a big hit on storage, and while it's true, but not always a bad thing. We will later look at why.
```

PR #2699, opened 2026-01-19
Add new string types to tables section of docs

atharvadeosthale:
## Summary by CodeRabbit

* **Documentation**
  * Introduced explicit string column types: varchar, text, mediumtext, longtext, with guidance on sizes and indexing.
  * Updated examples and SDK snippets to use varchar in place of the legacy string type.
  * Added example column definitions for bio (text), content (mediumtext), and data (longtext) with optional/required usage.
  * Added a deprecation notice for the legacy string column type and refreshed related examples.


src/routes/docs/products/databases/tables/+page.markdoc
```diff
@@ -1239,6 +1242,11 @@ You can choose between the following types.
 | `line`       | Geographic line represented by an ordered list of coordinates. |
 | `polygon`    | Geographic polygon representing a closed area; supports interior holes. |
 | `relationship` | Relationship column relates one table to another. [Learn more about relationships.](/docs/products/databases/relationships) |
+| `string`     | **Deprecated.** Use `varchar`, `text`, `mediumtext`, or `longtext` instead. |
+
+{% info title="Migrating from string columns" %}
+The `string` type is deprecated. Internally, `string` columns dynamically switch between `varchar`, `text`, `mediumtext`, and `longtext` based on the size you specify. Existing `string` columns will change to the appropriate text type based on the size of the existing column.
```
abnegate: > Existing `string` columns will change to the appropriate text type based on the size of the existing column.

This part is not true, existing columns won't be migrated automatically

src/routes/docs/products/databases/tables/+page.markdoc
```diff
@@ -48,7 +48,7 @@ const promise = tablesDB.createTable({
         },
         {
             key: 'name',
-            type: 'string',
+            type: 'varchar',
```
adityaoberai, line 51: @atharvadeosthale let's add examples of the other new types too (they don't mandatorily require a size afaik but let's double check)

src/routes/blog/post/new-string-types/+page.markdoc
```diff
@@ -0,0 +1,111 @@
...
+description: Appwrite Databases now supports varchar, text, mediumtext, and longtext column types, giving you explicit control over how string data is stored and indexed. The legacy string type is now deprecated.
+date: 2026-02-10
+cover: /images/blog/announcing-screenshots-api/cover.png
+timeToRead: 5
+author: atharva
+category: announcement
+featured: false
+---
+
+Until now, Appwrite offered a single `string` column type that internally switched between four different storage types based on the size you specified. This meant you had no visibility into how your data was actually stored, how it could be indexed, or why certain size limits existed.
```
eldadfux: I would use abstracted instead of switch or in addition

src/routes/blog/post/new-string-types/+page.markdoc
```diff
@@ -0,0 +1,111 @@
...
+
+`text` is stored off-page, with only a 20-byte pointer kept in the table row. This means it doesn't eat into your row size budget, but it can only be indexed with prefix or functional indexes. Use `text` when you need string storage without full indexing, like descriptions or notes.
+
+## Mediumtext and Longtext
+
+For larger content, `mediumtext` supports up to ~4 million characters and `longtext` up to ~1 billion characters. Both are stored off-page and support prefix indexing only. These are the right choice for storing things like article bodies, logs, or serialized data.
+
+# Why we made this change
+
+We regularly dogfood Appwrite in our own projects to find areas we can improve. While building [Imagine](https://imagine.dev), our AI-powered builder for web apps, which uses Appwrite under the hood for out-of-the-box server functionality, we quickly noticed how hard it is for AI models to reason about the `string` column type. Here's what we found.
```
eldadfux: We can share the Imagine story, but I think this confusion came from humans engineers before agents struggled with it. So we made the fix for both - win win

src/routes/blog/post/new-string-types/+page.markdoc
```diff
@@ -0,0 +1,111 @@
...
+            type: 'longtext',
+            required: false
+        }
+    ]
+});
+```
+
+# Deprecation of the string type
+
+The `string` column type is now deprecated. It will continue to work for existing columns, but we recommend using the explicit types for all new columns going forward.
```
eldadfux: I would make it clearer that we plan to continue backward support just to avoid any confusion even if its clear, we can make x2 clear.

src/routes/blog/post/new-string-types/+page.markdoc
```diff
@@ -0,0 +1,111 @@
...
+- **Not just an AI problem.** We've been observing this same issue in the community. Developers run into row width limits without understanding why, because the storage behavior of `string` was invisible.
+
+We saw this firsthand with Imagine. Imagine runs an Appwrite Cloud agent that provisions database resources on your behalf using the Appwrite CLI and `appwrite.json` as the schema. The agent is trained to build sophisticated architectures, and it would easily create just enough string columns to hit the row width limit.
+
+We could have solved this with better prompting, but that would only fix Imagine. We wanted to improve the experience for everyone, whether you're using AI tooling, the Console, or the SDKs directly. With explicit types, an AI model now naturally picks `longtext` for something like blog content, which is stored off-page and doesn't consume row space.
```
eldadfux: I'm missing something about the lack of visibility of the varchar row size limit, I think this was the root issue with the first approach. Now we make things more visible (even on the console) and help devs and agents alike make better decisions around their desired data structures.

eldadfux: Imagine use cases is great, but the human use case should get here equal attention. It should be 50:50

src/routes/blog/post/new-string-types/+page.markdoc
```diff
@@ -0,0 +1,109 @@
...
+
+`text` is stored off-page, with only a 20-byte pointer kept in the table row. This means it doesn't eat into your row size budget, but it can only be indexed with prefix or functional indexes. Use `text` when you need string storage without full indexing, like descriptions or notes.
+
+## Mediumtext and Longtext
+
+For larger content, `mediumtext` supports up to ~4 million characters and `longtext` up to ~1 billion characters. Both are stored off-page and support prefix indexing only. These are the right choice for storing things like article bodies, logs, or serialized data.
+
+# Why we made this change
+
+The root issue with the legacy `string` type was a lack of visibility. It abstracted away important storage details, which meant developers had no way to make informed decisions about their data structures. This led to problems for both human engineers and AI agents alike.
```
eldadfux: I would maybe add a bit more context to why we made this decision from the first place and be honest about us realizing that it compromised on important details like visibility and enough control to devs. We can also mention that finding the right abstraction balance is a difficult task, and from time to time based on user feedback and experience we have to adjust and respond which is what were doing now.

PR #2715, opened 2026-01-26
Copy prompt from a remote folder in docs

atharvadeosthale:


src/lib/preprocessors/promptAsContent.js
```diff
@@ -0,0 +1,68 @@
...
+            // Only process .markdoc files
+            if (!filename?.endsWith('.markdoc')) {
+                return;
+            }
+
+            // Check for promptAsContent in frontmatter
+            const frontmatterMatch = content.match(/^---\n([\s\S]*?)\n---/);
+            if (!frontmatterMatch) {
+                return;
+            }
```
ItzNotABug, line 27: do we not have a frontmatter parser somewhere already? 👀

src/lib/preprocessors/promptAsContent.js
TorstenDittmann, line 1: I would like to avoid adding a preprocessor. Only if it would affect 90% of the markdown files it would be suitable.

cant we just pass the routeId as frontmatter to https://github.com/appwrite/website/blob/9c8868b80d84ae2945f6a4a1732ce8c9091b298e/src/lib/utils/routePrompts.ts#L27 ? that way we can use the existing file without having to copy it.

atharvadeosthale: I tried having an implementation where it just uses a markdown (prompt.md) file at a route. The problem with that is that we lose all styling because it uses some other markdown processor (I think it-markdown). I checked how partials are processed, and it's through another pre-processor. Since I couldn't go around any other way, I decided to do this. Lmk ur thoughts @TorstenDittmann

PR #2883, opened 2026-04-10
Avoid duplicated metadescriptions

DH-555:
## What does this PR do?

Uses custom descriptions for those pages instead of using the default metadescription.
Using the same is bad for SEO and not a good practice.


eldadfux:

PR #2948, opened 2026-04-30
perf(marketing): client-side render below-the-fold homepage sections

TorstenDittmann:
## Summary
- Client-side renders all homepage sections below LogoList (Bento, Pullquote, Ai, CaseStudies, Features, Map, Scale, Pricing) to eliminate ~580ms of SSR CPU time per request.
- Adds a single `100vh` SSR placeholder to keep the footer below the fold on initial paint.
- Keeps Hero, Platforms, LogoList, FooterNav, and MainFooter SSR'd for fast first paint and SEO coverage.
- Removes an unused `browser` import from `logo-list.svelte`.

## Expected impact
| Metric | Before | After |
|--------|--------|-------|
| SSR time (warm) | ~585 ms | ~5 ms |
| Speedup | — | ~100× |

## Notes
- `prerender = false` is intentionally preserved so Statsig hero experiments continue to be evaluated server-side.
- This is a band-aid fix; the long-term solution is prerendering with client-side experiment assignment and designed skeleton states.

### DIFF
--- a/src/routes/(marketing)/(components)/logo-list.svelte
+++ b/src/routes/(marketing)/(components)/logo-list.svelte
@@ -1,5 +1,4 @@
 <script lang="ts">
-    import { browser } from '$app/environment';
     import { cn } from '$lib/utils/cn';
 
     type Props = {
--- a/src/routes/(marketing)/+page.svelte
+++ b/src/routes/(marketing)/+page.svelte
@@ -34,8 +34,8 @@
             title="App Manager, Majik Kids"
             avatar="/images/testimonials/phil.jpg"
         >
-            <span class="text-secondary">Just like a Swiss Army Knife</span> you can choose and use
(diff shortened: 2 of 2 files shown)


eldadfux (approved):

TorstenDittmann (commented):
@greptile-apps Noted on the SEO concern, but this is an intentional trade-off for this specific page. The marketing homepage relies on server-side Statsig experiment evaluation for the hero section, which requires `prerender = false`. The SEO-critical content (hero copy, platform descriptions, footer links) remains SSR'd. The below-the-fold sections (features, pricing, etc.) are not primary landing traffic drivers — users land on dedicated `/products/*` and `/pricing` pages for that content. This is documented as a band-aid fix until we can prerender with client-side experiment assignment.

PR #3081, opened 2026-07-03
ci: write image tag to global.image.tag

levivannoort:
The applications repo moves `image` to `global.image` (generic chart 1.7.0 — see appwrite-labs/helm-charts#17); this updates the tag-bump automation to write the single canonical path `.global.image.tag` in `default.yaml`.

**Merge order**: land appwrite-labs/assets-applications#29 first. Until then, an old-style per-alias write still overrides `global`, so deploys stay correct in either interim state (a stale per-alias tag key would just need removing).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/production.yml
+++ b/.github/workflows/production.yml
@@ -86,7 +86,7 @@ jobs:
                   token: ${{ steps.app-token.outputs.token }}
 
             - name: Update image tag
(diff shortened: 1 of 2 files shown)


ChiragAgg5k (approved):
Reviewed the deployment workflow updates. Both environments now write the image tag to the global chart path while keeping the existing commit flow intact. Looks good.

PR #3113, opened 2026-07-21
docs(databases): describe HA replication honestly as async-only

abnegate:
## Problem (DAT-1942)

nyc3 2026-07-21 Drill C proved `syncMode` is stored-but-inert: `sync`/`quorum` were accepted and echoed while every engine replicated async (commits under `syncMode=sync` with the replica scaled down completed in milliseconds; no `synchronous_standby_names`, no semi-sync plugin, no quorum handling). The HA docs page promised exactly that missing durability:

- The sync-modes table: `sync` = "at least one replica acknowledges", `quorum` = "a majority acknowledge" — plus "a safe default for production is `sync`".
- The engine table: MySQL/MariaDB "Semi-synchronous binlog replication".
- Read-your-writes: "Use a stronger sync mode … any reader sees the write as soon as COMMIT returns".
- Every code sample requested `'sync'` — which the API now rejects (appwrite-labs/edge#890, appwrite-labs/cloud#4894).

## Fix (minimal, claims only)

- Sync-modes section now states replication is asynchronous, `async` is the only accepted value (`sync`/`quorum` rejected until implemented), with the real crash-loss disclosure semantics.
- Engine table rows say asynchronous.
- Read-your-writes third bullet now gives the guidance that is actually wired: reads inside a transaction or `SELECT … FOR UPDATE` pin to the primary via the pooler (verified in the pooler routing rules, already documented on the pooler page — that page needed no changes).
- Code samples request `async`; dropped the "sync mode change online" limits row; frontmatter description no longer advertises synchronous/semi-synchronous.

Base is `feat-dedicated-db` (the branch carrying these docs, PR #3018), matching #3042's stacking.

Refs DAT-1942.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/routes/docs/products/databases/dedicated/high-availability/+page.markdoc
+++ b/src/routes/docs/products/databases/dedicated/high-availability/+page.markdoc
@@ -1,7 +1,7 @@
 ---
 layout: article
 title: High availability
-description: Run up to five replicas of a dedicated database with synchronous, semi-synchronous, or asynchronous replication and automatic failover on primary failure.
+description: Run up to five replicas of a dedicated database with asynchronous replication and automatic failover on primary failure.
 ---
 
 High availability adds replicas to a dedicated database and enables automatic failover when the primary becomes unhealthy. With HA on, a primary outage is recovered in seconds by promoting the most up-to-date replica, without manual intervention.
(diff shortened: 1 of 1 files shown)


abnegate:
Closing per Jake's direction — the honest-async-only docs approach is being dropped in favor of a different direction on syncMode.

PR #3171, opened 2026-08-14
Add Appwrite and Vercel MCP benchmark post

ChiragAgg5k:
## Summary

- Add a write-enabled Appwrite MCP and Vercel MCP benchmark post, matching the original write-up at https://www.chiragaggarwal.tech/blog/appwrite-mcp-vs-vercel-mcp word for word.
- Compare deployment, observability, latency, safety, rollback, and tool coverage: Appwrite reaches further into the backend, Vercel deploys faster and filters logs better.
- No combined verdict. The two totals landed too close to mean anything, so the post reports the per-task scores that were actually recorded instead.

```text
Appwrite MCP → 992 operations across 81 services, behind 4 meta-tools
Vercel MCP   → 33 named tools, called directly
```

## Figures

Four figures carry the data, each composed at 16:9 because the blog crops post images to that ratio.

### Run sheet

The six benchmark stages and where each server pulled ahead

### Per-task scores

Per-task scores out of 5 across the eight read and observability tasks

### Tool reach

33 named Vercel tools against 992 Appwrite operations, drawn to the same scale

### Observed latency

Observed MCP call latency in milliseconds across six repeated reads

## Testing

```text
bun run dev
```

Rendered the post locally and confirmed all five AVIF assets load at their intended resolution and stay legible inside the blog's 16:9 image frame.


adityaoberai:
## Blog review: `we-benchmarked-appwrite-mcp-against-vercel-mcp`

Reviewed against our blog standards (voice, structure, formatting, SEO, accuracy) plus a fact-check of every numeric claim against the linked report.

Structurally and tonally this is one of the stronger posts we've shipped: no throat-clearing opener, **zero em dashes**, declarative section headings that commit to verdicts, and unusually disciplined caveats for a vendor-authored competitor comparison. Two things need to happen before publish, and SEO needs a pass.

### 1. The headline score contradicts the report it cites

The post says the result was a tie with **both servers at 4.1/5**. The linked report says **Appwrite 4.2/5, Vercel 4.1/5**. A reader who clicks through, and we push that link three times, sees a different number in the first screenful.

This needs fixing in three places:
- body, the lede paragraph ("The result was a tie…")
- the FAQ answer ("scored both servers 4.1/5, a directional tie")
- **`cover.avif` itself**, which renders `4.1 / 5` on both cards, so this is an image re-export, not just a text edit

The "tie" framing then cascades into the whole "different jobs, not better/worse" structure. Worth noting the error runs *against* us, so fixing it isn't a fairness problem:

```diff
- The result was a tie: **Appwrite and Vercel both scored 4.1/5**. The more useful finding is that the servers are built for different jobs.
+ The scores landed close: **Appwrite 4.2/5, Vercel 4.1/5**. A 0.1 gap is a directional result, not a verdict, and the more useful finding is that the servers are built for different jobs.
```

### 2. The report's most severe finding, an Appwrite bug, is omitted

From the report:

> Successful runtime logs labeled as failure — `sites_list_logs` returned a successful raw payload with 1,417 total executions and 20 detailed request events, but SDK model deserialization failed and the MCP returned `ok:false`.

Appwrite scored **2.4/5** on runtime logs against Vercel's **4.2/5**. The post renders this as:

> Both platforms handled the requests correctly. […] Appwrite returned detailed request data including status, duration, stdout, and stderr.

That describes the raw payload while omitting that the server told the agent it had failed. It's the one paragraph that undercuts the credibility the rest of the post works hard to earn, and it's prominent in the report we link three times. Suggested rewrite:

```diff
- The runtime test generated three 200 responses, one 400, and one 500. Both platforms handled the requests correctly. Vercel's dedicated runtime tools returned the five events and grouped the 500 into a route-specific error cluster. Appwrite returned detailed request data including status, duration, stdout, and stderr.
+ The runtime test generated three 200 responses, one 400, and one 500. Both platforms served them correctly, but the MCP surfaces did not report them equally well. Vercel's dedicated runtime tools returned the five events and grouped the 500 into a route-specific error cluster. Appwrite's `sites_list_logs` fetched the data successfully, including 1,417 executions and 20 detailed request events, then failed to deserialize the payload and returned `ok:false` to the agent. The data was there; the tool reported failure anyway. Runtime logs were Appwrite's weakest category at 2.4/5 against Vercel's 4.2/5, and the deserialization bug is the fix we took away from this run.
```

### 3. SEO, the weakest checklist area

| Check | Target | Actual |
|---|---|---|
| `title` length | 50-60 chars | **46** |
| `description` length | 150-160 chars | **131** |
| `timeToRead` | ~200-250 wpm | **8** for 1,564 words (~6) |
| `unlisted` present | required | **absent** |
| Contextual in-body internal links | required | **zero** |
| Appwrite Cloud link in closing | required | **absent** |
| Body word count | 1,500-2,000 | 1,564 ✅ |

```diff
- title: "We benchmarked Appwrite MCP against Vercel MCP"
+ title: "Appwrite MCP vs Vercel MCP: we benchmarked both"

- description: A hands-on comparison of Appwrite and Vercel MCP across deployment, latency, observability, rollback, safety, and failure recovery.
+ description: We benchmarked Appwrite MCP vs Vercel MCP across 12 tests covering deployment speed, latency, logs, rollback, and write safety. Here is what each server won.

- timeToRead: 8
+ timeToRead: 6
  featured: false
+ unlisted: false
```

`against` to `vs` matters: that's what people actually type. Same reasoning applies to the slug, and since this is unpublished, `appwrite-mcp-vs-vercel-mcp-benchmark` is still available and front-loads the keyword.

The body currently has **no contextual internal links at all**. The only internal link in the post is `/docs/tooling/ai/mcp-servers/api` in the closing list. Three directly relevant targets exist:
- `/docs/products/sites/instant-rollbacks`, link from "rolled back to v1 in 1.32 seconds"
- `/docs/products/sites/logs`, link from the logs section
- `/docs/products/sites/deployments`, link from the activation discussion

Closing section also needs the Cloud link, and the Discord link is filler here:

```diff
- - [Appwrite Discord community](https://appwrite.io/discord)
+ - [Instant rollbacks for Appwrite Sites](/docs/products/sites/instant-rollbacks)
+ - [Start building on Appwrite Cloud](https://cloud.appwrite.io)
```

### 4. Smaller fixes

**"Platform-specific reasons" hides the best detail in the failure test.** The two reasons were asymmetric and both worth naming, and naming Vercel's security check is good for our even-handedness:

```diff
- Both platforms correctly rejected an earlier version of the fixture for platform-specific reasons. After adding `next.config.js` and upgrading Next.js from 15.2.4 to 15.2.8, the same corrected source deployed successfully to both.
+ Both platforms rejected an earlier version of the fixture, for different and legitimate reasons. Appwrite's adapter compiled the build, then failed because it expected a `next.config.*` file. Vercel completed the build, then blocked it because Next.js 15.2.4 carried a known vulnerability. Adding `next.config.js` and upgrading to 15.2.8 fixed both, and each agent diagnosed its own failure correctly from the logs it was given.
```

**"About 2.3 times faster" is ambiguous.** Two Appwrite numbers precede it (91s build, 121s total). 91/39.1 = 2.33 but 121/39.1 = 3.09, so a reader checking the math gets a different answer:

```diff
- On this fixture, Vercel's valid SSR build was about 2.3 times faster.
+ Comparing reported build time alone, Vercel's valid SSR build was about 2.3 times faster on this fixture.
```

**Question-format heading** (banned by our checklist), and one heading is internal jargon nobody searches:

```diff
- # Which MCP server should you use?
+ # Choosing between Appwrite MCP and Vercel MCP

- # Router versus toolbox
+ # Tool discovery: searchable router versus direct toolbox
```

**Missing content:**
- **No per-category score breakdown.** We give the six rubric weights (they sum to 100% ✅) and a composite, but never the sub-scores, so the headline number can't be checked without leaving the page. The report has them: workspace discovery 3.4/4.6, build logs 4.0/4.6, runtime logs 2.4/4.2, usage/analytics 4.7/3.7.
- **The recovery stage result is never stated**, even though error recovery is 10% of the score and stage 5 of our own method diagram. The body only covers log *quality*. Per the report, both agents recovered.

**Structure:** all nine headings are H1, no H2/H3 anywhere. The three longest sections would scan much better broken up. The logs section alone covers build logs, runtime logs, and docs search as one undifferentiated run.

**Nits:** straighten the curly quotes in `“Not exposed”` (straight quotes everywhere else in the post); `version:"v1"` is missing a space; "faster on several repeated reads" understates it, Appwrite was faster on all five non-orientation operations; the closing heading "Try Appwrite MCP" is boilerplate rather than specific to this post.

### 5. Two things to check outside the post

- **`confirm_write=true` appears nowhere in `src/routes/docs/`.** This post is currently its only documentation, and it's load-bearing for the whole write-safety argument. Either confirm the parameter name against the server or file a docs follow-up.
- **The load-bearing citation sits on a throwaway host.** Every number traces to `mcp-platform-report-0813.appwrite.network`, a dated, preview-shaped subdomain. If that deployment is ever cleaned up, the post's entire evidence base 404s. Worth a permanent home.
- Minor: the cover uses Vercel's triangle mark in Appwrite marketing artwork. Probably fine as nominative use, but easier to get a sign-off now than after publish.

### Verified clean

Frontmatter matches the `PostsData` contract in `src/routes/blog/content.ts`; author `chirag-aggarwal` exists; `category: ai, comparisons` follows the multi-category convention and both category pages exist; `faqs` is supported by `Post.svelte` and markdown-in-answers is normal (657 of 3,114 existing answers contain links); `#`-for-sections matches convention; the `.avif` cover matches all 561 posts.

All four images are 1280x720, visually consistent, and accurate to the body. The latency chart's bars match the table, its "single successful call" annotations match the caveat in prose, and both charts carry their own caveat footers. Alt text is descriptive throughout.

Every Appwrite claim checks out against local docs: `mcp.appwrite.io` and OAuth (`tooling/ai/mcp-servers/api`), all four meta-tool names verbatim, the activation/rollback model (`products/sites/deployments`, where a `ready` deployment can be activated to replace the active one), and product naming (Auth, Databases, Storage, Messaging, Functions, Sites all match doc page titles).

Every number except the score matches the report: 33 tools, 992 operations across 81 services, 2.1s, ~27s, 39.1s, 91s, ~121s, 1.32s, 5.146s vs 4.714s, exit code 42, and the weight table. The Vercel gaps are accurate too. The report confirms "No rollback/redeploy-to-existing-project tool exposed" and "No project deletion tool is exposed in the 33-tool MCP."

The caveat discipline throughout ("'Not exposed' refers only to the MCP tool set we tested", "not a controlled infrastructure or load test", "too small to support a broader performance claim") is genuinely good, and it's what makes items 1 and 2 worth correcting rather than letting slide.

### Priority

1. Correct the score in all three places, including the cover re-export.
2. Report the `sites_list_logs` bug and the 2.4/5 runtime-logs score.
3. Fix the SEO basics: title, description, in-body doc links, Cloud link.

Then: name the two fixture-rejection reasons, add the per-category score table, state the recovery outcome, break the long sections with H2s.

adityaoberai:
## Follow-up: the 4.2 vs 4.1 composite doesn't reconcile, and my earlier table was a bad cherry-pick

Two corrections to my review above, the second more serious than the first.

### 1. Correcting myself: the task-score table I suggested was a cherry-pick

In my review I suggested adding a per-task score table and quoted four rows: workspace discovery 3.4/4.6, build logs 4.0/4.6, runtime logs 2.4/4.2, usage/analytics 4.7/3.7.

The report's table actually has **eight** rows. The four I quoted contained three of Appwrite's four losses, which gives exactly the wrong impression. That's the same selective framing I flagged the post for, so it's worth correcting in public. The full table:

| Task | Appwrite | Vercel | Winner |
|---|---:|---:|---|
| Workspace and project discovery | 3.4 | 4.6 | Vercel |
| Resource detail | 4.6 | 4.4 | Appwrite |
| List deployments | 3.8 | 3.5 | Appwrite |
| Deployment detail | 4.6 | 4.3 | Appwrite |
| Build logs | 4.0 | 4.6 | Vercel |
| Runtime logs | 2.4 | 4.2 | Vercel |
| Usage and analytics | 4.7 | 3.7 | Appwrite |
| Documentation search | 4.3 | 3.5 | Appwrite |

Appwrite wins **5 of 8**, not 1 of 4.

### 2. The bigger problem: the combined score can't be derived from anything published

Even with all eight rows, the arithmetic runs against the headline:

- Appwrite mean: **3.97**
- Vercel mean: **4.10**

Appwrite wins more tasks but loses the average, because its three losses are heavier (-1.2, -0.6, -1.8) than most of its wins. Read straight, this table suggests Vercel is ahead, yet the post and report both report Appwrite 4.2 vs Vercel 4.1.

I went back through the report looking for the reconciliation and it isn't there. The report publishes:

- the six rubric categories and their weights
- the eight task scores above
- the two combined totals

It does not publish numeric scores for any of the six weighted categories, nor any scores for the write, rollback, cleanup, safety, or concurrency phases. The methodology section describes a JSON schema for capturing scores but never displays populated results.

Those unpublished categories are **55% of the total weight** (correctness and completeness 30%, safety and lifecycle 15%, error recovery 10%), and they're exactly where Appwrite's rollback, deletion, and `confirm_write` advantages would land. So there's a very plausible story for how the composite flips. But right now it is a story, not published arithmetic.

This matters more than a normal sourcing gap because of who is publishing. A number that cannot be checked, in a comparison written by one of the two vendors, that lands 0.1 in the publisher's favour, is the weakest possible thing to lead with. It also sits awkwardly against the post's own opening line, that an MCP server should be judged by what an agent can complete rather than by a single figure.

### The ask

@ChiragAgg5k do you still have the populated scoring data from the run? Specifically the six weighted category scores per platform:

| Category | Weight | Appwrite | Vercel |
|---|---:|---:|---:|
| Correctness and completeness | 30% | ? | ? |
| Result compactness | 15% | ? | ? |
| Discoverability and call economy | 15% | ? | ? |
| Median and p95 latency | 15% | ? | ? |
| Safety and lifecycle control | 15% | ? | ? |
| Error recovery quality | 10% | ? | ? |
| **Combined** | | **4.2** | **4.1** |

If those exist, publishing them in the report and adding the table to the post makes the headline auditable and the whole piece much harder to argue with. If they don't exist in that form, the cleaner option is to drop the composite from the lede, the FAQ, and the cover, and lead on the dimensional result instead: Vercel deployed faster and reported runtime failures better, Appwrite reached further into the backend and was the only one exposing rollback, deletion, and write confirmation. That version needs no data we don't have, and it's closer to the post's own thesis anyway.

Flagging that this decision also affects the cover re-export already needed for the score correction, so it's worth settling before that artwork gets redone rather than after.

### Status of the other fixes

The rest of the review is applied locally and waiting on this decision before it goes up: frontmatter (title 55 chars, description 157, `unlisted` added), the score correction in lede and FAQ, the runtime-log bug and 2.4/5 disclosure, the two real fixture-rejection reasons, the recovery outcome, H2 subsections, in-body doc links, Cloud link in, Discord filler out, and the smaller nits. The task-score table in that local copy now carries all eight rows plus a note that they cover the read and observability phases only and do not average to the combined score.

`timeToRead` is worth a note: I initially dropped it to 6 for the original 1,564 words, but the added disclosure and structure took the body to ~1,976, so it goes back to **8**. The original value was right for the final length.

ChiragAgg5k:
Looked into the one finding (P2, missing `.optimize-cache.json` entries for the five AVIFs) and it doesn't apply here.

`.optimize-cache.json` is keyed by the **pre-conversion raster source**, not by the AVIF output. `scripts/optimize-assets.js` only walks `RASTER_EXT = ['.png', '.jpg', '.jpeg', '.webp']`, so `.avif` files are never visited and never hashed. Of the 2,304 keys in the file today, 2,299 are `.png`/`.jpg`/`.jpeg`/`.webp`; the two `.avif` keys are legacy leftovers rather than the convention. These five images were authored as AVIF and committed directly, so there is no source raster to record and nothing for a future optimization run to recognize.

Verified rather than assumed — a full local run is a no-op with a clean tree:

```
$ bun ./scripts/optimize-assets.js
Done. Converted 0 file(s) this run.
$ git status --short
(empty)
```

The repo's own gate agrees: the `assets` job runs `CI=true bun run optimize` and fails on any resulting diff, and it passes on this branch.

No change made for this one. Everything else on the PR is green.

PR #3173, opened 2026-08-14
feat(docs): add custom MFA channel guides for Telegram and WhatsApp

atharvadeosthale:
Adds a guide that shows how to deliver the custom MFA code through a real channel. The [custom MFA factor](https://github.com/appwrite/website/pull/3159) page covers the flow. This page covers the two parts that page leaves to the reader: where the destination comes from, and what the provider call looks like.

One page, `/docs/products/auth/custom-mfa-channels`, with a tab per channel across three steps.

**Telegram**

- Bot setup through BotFather.
- Enrollment through a one-time deep link, so the chat ID is bound to the Appwrite user instead of taken from the client.
- A webhook function that receives the chat ID, with the secret header check.
- The delivery call.

**WhatsApp**

- Meta app setup and authentication template creation.
- The destination is the phone number already in the Appwrite user record, gated on `phoneVerification`.
- The delivery call, including the OTP button component.

**Verification**

Every Appwrite call, scope name, and response shape was checked against SDKs generated from the build the custom factor landed in. Both provider payloads were checked against provider documentation. Neither channel has been executed end to end yet: Telegram is testable on demand, WhatsApp needs Meta template approval first.

**Notes for review**

- Based on `atharva/custom-mfa-factor-docs`, because the page links to the custom MFA factor page and sits below it in the sidebar. Merge that one first.
- Node only. The channel call is plain HTTP, so the runtime is not the subject of this page, and the Rust equivalent of the read step already appears on the custom MFA factor page.

### DIFF
--- a/src/routes/docs/products/auth/+layout.svelte
+++ b/src/routes/docs/products/auth/+layout.svelte
@@ -132,6 +132,11 @@
                     new: isNewUntil('30 November 2026')
                 },
                 {
+                    label: 'Custom MFA channels',
+                    href: '/docs/products/auth/custom-mfa-channels',
+                    new: isNewUntil('30 November 2026')
+                },
+                {
                     label: 'Auth status check',
                     href: '/docs/products/auth/checking-auth-status'
                 },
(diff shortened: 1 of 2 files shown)


eldadfux:

PR #3191, opened 2026-08-21
fix: support standard OpenAPI example fields in refs

atharva-appwrite:
## Summary
- update the reference spec loader to read standard OpenAPI `example` fields for JSON and multipart request-body properties
- keep `x-example` as a fallback so older specs still render examples
- align website reference rendering with the sdk-generator change in PR #1812

## Verification
- verified the website code path in `src/routes/docs/references/[version]/[platform]/[service]/specs.ts`
- verified sdk-generator PR #1812 switched to standard OpenAPI example fields
- `bun run check` was not run in this clean clone because the repo dependencies are not installed here


eldadfux:

